PR #8, opened 2022-12-01
Update location.yaml

ukkaps:
Included request parameter “MaxLocationAge’ to give the developer option to mention the maximum aging of the location info, accepted by the application. 

If the developer wants only real-time or near real-time information, we as operators then can avoid validating against stale location info resulted from network reasons such as device switched-off, not in servicing area, device issues etc.

### DIFF
--- a/code/API_definitions/location.yaml
+++ b/code/API_definitions/location.yaml
@@ -97,14 +97,14 @@ components:
           $ref: '#/components/schemas/Longitude'
         accuracy:
           $ref: '#/components/schemas/Accuracy'
-        maxLocationAge:
-          $ref: '#/components/schemas/MaxLocationAge'
+        maxAge:
+          $ref: '#/components/schemas/MaxAge'
       required:
         - ueId
         - latitude
         - longitude
         - accuracy
-        - maxLocationAge 
+        - maxAge 
     VerifyLocationResponse:
       type: object
       required:
@@ -184,7 +184,7 @@ components:
       minimum: 2
       maximum: 200
       example: 50
-    MaxLocationAge:
+    MaxAge:
       description: Max location aging validity required by the application in Minute, Default =0 for realtime location
       type: number
       minimum: 0
--- a/code/API_definitions/location.yaml
+++ b/code/API_definitions/location.yaml
@@ -8,7 +8,7 @@ info:
   license:
     name: Apache 2.0
     url: https://www.apache.org/licenses/LICENSE-2.0.html
-  version: 0.1.0
+  version: 0.2.0
 externalDocs:
   description: Product documentation at Camara
   url: https://github.com/camaraproject/
--- a/code/API_definitions/location.yaml
+++ b/code/API_definitions/location.yaml
@@ -97,11 +97,14 @@ components:
           $ref: '#/components/schemas/Longitude'
         accuracy:
           $ref: '#/components/schemas/Accuracy'
+        maxLocationAge:
+          $ref: '#/components/schemas/MaxLocationAge'
       required:
         - ueId
         - latitude
         - longitude
         - accuracy
+        - maxLocationAge 
     VerifyLocationResponse:
       type: object
       required:
@@ -181,6 +184,12 @@ components:
       minimum: 2
       maximum: 200
       example: 50
+    MaxLocationAge:
+      description: Max location aging validity required by the application in Minute
+      type: number
+      minimum: 1
+      maximum: 1440
+      example: 30
     VerificationResult:
       description: Result of a verification request, true on match
       type: boolean
--- a/code/API_definitions/location.yaml
+++ b/code/API_definitions/location.yaml
@@ -185,9 +185,9 @@ components:
       maximum: 200
       example: 50
     MaxLocationAge:
-      description: Max location aging validity required by the application in Minute
+      description: Max location aging validity required by the application in Minute, Default =0 for realtime location
       type: number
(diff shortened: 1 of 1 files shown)


code/API_definitions/location.yaml
```diff
@@ -97,11 +97,14 @@ components:
...
         accuracy:
           $ref: '#/components/schemas/Accuracy'
+        maxLocationAge:
+          $ref: '#/components/schemas/MaxLocationAge'
       required:
         - ueId
         - latitude
         - longitude
         - accuracy
+        - maxLocationAge
```
jlurien, line 107: Maybe it should be optional, with a default value. It depends on how often it is expected to be used. If the common case is to request "real-time" info (less than a minute or few minutes), it is not convenient to ask for it in every request.

jlurien (commented):
It is a good addition as long as it is expected that some implementations may rely on stale data. In TEF we were thinking on a "real-time" implementation, so this input is not really required, but I would like to hear opinions from other operators. If convenient, we think it would be better to leave it optional with a default value to the minimum. We also need to define the expected behaviour when the location is known but the age threshold is not met. Should API return a response with a warning, an error, a false response...?

jlurien:
We also have to review the versioning process. Any modification should increase current version (0.1.0). We may need to create a new development branch towards 0.2.0

bigludo7:
Hello
This is a valuable request. We need on our side a little bit more time to explore 3GPP api capabilities (which support the camara api implementation on our side). 

Agreed on @jlurien comments on versioning - as we have already begun implementation it's better to change swagger version

patrice-conil:
Hello @ukkaps @jlurien ,
As MaxLocationAge = 0  can  mean for "current" or "real time" ... I vote for mandatory version.

Patrice

ukkaps:
@bigludo7  @patrice-conil  @jlurien 
API version updated as suggested from 0.1.0 to 0.2.0

code/API_definitions/location.yaml
patrice-conil: A default value for maxLocationAge is a good idea. I vote for.

code/API_definitions/location.yaml
```diff
@@ -181,6 +184,12 @@ components:
       minimum: 2
       maximum: 200
       example: 50
+    MaxLocationAge:
+      description: Max location aging validity required by the application in Minute
+      type: number
+      minimum: 1
+      maximum: 1440
+      example: 30
```
patrice-conil, line 192: default: 0

patrice-conil (changes requested):
Concerning Ipv6Addr. 
As the second pattern is much too permissive and does not add anything more, I suggest deleting it.

    Ipv6Addr:
      type: string
      format: ipv6
      pattern: '^((:|(0?|([1-9a-f][0-9a-f]{0,3}))):)((0?|([1-9a-f][0-9a-f]{0,3})):){0,6}(:|(0?|([1-9a-f][0-9a-f]{0,3})))(\/(([0-9])|([0-9]{2})|(1[0-1][0-9])|(12[0-8])))?$'
      example: '2001:db8:85a3:8d3:1319:8a2e:370:7344'

code/API_definitions/location.yaml
jlurien, line 107:
```suggestion
        - maxAge 
```
Location is implicit in the context

code/API_definitions/location.yaml
jlurien: I have to check the technical implications, but I assume that maxAge=0 implies that implementation must not rely on any kind of stale or cached data. If we set the default to this, we are forcing that requests which not include explicitly this parameter would get an error, if some implementation is caching some data for few minutes, which could not be the intention of the consumer.

jlurien:
Hi, in order to manage versioning and keep current agreed version in main, I have copy what we are doing in QualityOnDemand WG, and I have created a new branch **dev-0.2.0** where we should propose any new PR for 0.2.0, so please @ukkaps could you switch your proposal to be merged in the new branch instead of main.

jlurien (commented):
Some comments on the behaviour and implications of the default value. Also, please, could you move this PR to the new dev-0.2.0 branch. We should have created the new branch in advance but it's better now that when new changes arrive.

ukkaps:
@jlurien , new PR is raised [here](https://github.com/camaraproject/DeviceLocation/pull/9) as suggested above. May I close the old PR (this) now?

MarkusKuemmerle:
> @jlurien , new PR is raised [here](https://github.com/camaraproject/DeviceLocation/pull/9) as suggested above. May I close the old PR (this) now?
> 
> @MarkusKuemmerle , do we have any timelines to release the code of this for MWC implementation?



> @jlurien , new PR is raised [here](https://github.com/camaraproject/DeviceLocation/pull/9) as suggested above. May I close the old PR (this) now?
> 
> @MarkusKuemmerle , do we have any timelines to release the code of this for MWC implementation?

The 0.1.0 version which was merged in main branch in November will be the version for the MWC demos. Nevertheless it's good to further improve the API and to work on further versions.

jlurien:
> @jlurien , new PR is raised [here](https://github.com/camaraproject/DeviceLocation/pull/9) as suggested above. May I close the old PR (this) now?
> 
> @MarkusKuemmerle , do we have any timelines to release the code of this for MWC implementation?

@ukkaps, yes, you can close this now and we can follow discussions in the new one. Thanks!

PR #9, opened 2022-12-29
Update location.yaml

ukkaps:
Included request parameter “MaxAge’ to give the developer option to mention the maximum aging of the location info, accepted by the application.

If the developer wants only real-time or near real-time information, we as operators then can avoid validating against stale location info resulted from network reasons such as device switched-off, not in servicing area, device issues etc.

### DIFF
--- a/code/API_definitions/location.yaml
+++ b/code/API_definitions/location.yaml
@@ -97,14 +97,14 @@ components:
           $ref: '#/components/schemas/Longitude'
         accuracy:
           $ref: '#/components/schemas/Accuracy'
-        maxLocationAge:
-          $ref: '#/components/schemas/MaxLocationAge'
+        maxAge:
+          $ref: '#/components/schemas/MaxAge'
       required:
         - ueId
         - latitude
         - longitude
         - accuracy
-        - maxLocationAge 
+        - maxAge 
     VerifyLocationResponse:
       type: object
       required:
@@ -184,7 +184,7 @@ components:
       minimum: 2
       maximum: 200
       example: 50
-    MaxLocationAge:
+    MaxAge:
       description: Max location aging validity required by the application in Minute, Default =0 for realtime location
       type: number
       minimum: 0
--- a/code/API_definitions/location.yaml
+++ b/code/API_definitions/location.yaml
@@ -8,7 +8,7 @@ info:
   license:
     name: Apache 2.0
     url: https://www.apache.org/licenses/LICENSE-2.0.html
-  version: 0.1.0
+  version: 0.2.0
 externalDocs:
   description: Product documentation at Camara
   url: https://github.com/camaraproject/
--- a/code/API_definitions/location.yaml
+++ b/code/API_definitions/location.yaml
@@ -97,11 +97,14 @@ components:
           $ref: '#/components/schemas/Longitude'
         accuracy:
           $ref: '#/components/schemas/Accuracy'
+        maxLocationAge:
+          $ref: '#/components/schemas/MaxLocationAge'
       required:
         - ueId
         - latitude
         - longitude
         - accuracy
+        - maxLocationAge 
     VerifyLocationResponse:
       type: object
       required:
@@ -181,6 +184,12 @@ components:
       minimum: 2
       maximum: 200
       example: 50
+    MaxLocationAge:
+      description: Max location aging validity required by the application in Minute
+      type: number
+      minimum: 1
+      maximum: 1440
+      example: 30
     VerificationResult:
       description: Result of a verification request, true on match
       type: boolean
--- a/code/API_definitions/location.yaml
+++ b/code/API_definitions/location.yaml
@@ -185,9 +185,9 @@ components:
       maximum: 200
       example: 50
     MaxLocationAge:
-      description: Max location aging validity required by the application in Minute
+      description: Max location aging validity required by the application in Minute, Default =0 for realtime location
       type: number
-      minimum: 1
+      minimum: 0
       maximum: 1440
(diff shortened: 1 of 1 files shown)


bigludo7 (approved):
Look good for me;
Thanks

code/API_definitions/location.yaml
```diff
@@ -181,6 +184,12 @@ components:
       minimum: 2
       maximum: 200
       example: 50
+    MaxAge:
+      description: Max location aging validity required by the application in Minute, Default =0 for realtime location
```
jlurien, line 188: We were discussing in the original PR whether this new parameter should be mandatory or not. If mandatory, it makes no sense to define or document a default value.

jlurien (commented):
In v0.1 this parameter does not exist, so it is like client does not care about location age, and any implementation of the API regarding location age works. In practice this is as having a default value to a maximum.

To me this parameter makes sense when the client wants to request a location age that is more restrictive than certain default value, as I doubt a client would explicitly request anything worse than default (unless it is cheaper but we are not considering those aspects yet). 

If we make it required, client is more conscious about the age concept but API may be more difficult to use as clients may doubt what value in minutes to ask, and they would ask for the minimum usually.

I think we have to know well what can be expected from the network in the common cases, and what can be assumed as "real-time".  For example, if networks cannot not easily assure ages as low as 0 o 1 minutes, allowing clients to request it may be problematic, as clients probably tend to request the minimum if they have to provide a value mandatorily, For most use cases, low values in the order of few minutes would make no difference, but strictly some implementation may not fulfill age=0.  

@bigludo7 do you have a strong preference on this optional vs mandatory decision? Maybe you can check also internally what a minimum "reasonable" age of data in your network is.

Also, in any case we should define also the expected behaviour when the requested age cannot be fulfilled.

bigludo7:
Hello @jlurien 
Regarding your question, my understanding is:

-  value 0 mean 'now' (real time localisation)
-  it should be the value default. 
- mandatory or not? I haven't a strong opinion but I understood that https://github.com/patrice-conil has a preference for the mandatory version.
- as we use 3GPP Monitoring event, the age could be useful when the location type is "LAST_KNOWN_LOCATION" (and not "CURRENT_LOCATION").  Indeed if I got from my back end system a localisation with a  "ageOfLocationInfo"= 19, if you asked me with maxAge = 0  I cannot localize you while if you asked me with maxAge = 20 I send back  "verificationResult": true. I'm not sure about the maximum ageOfLocalisation we are able to manage in the BE.
- Completely agree with your point about defining also the expected behavior when the requested age cannot be fulfilled. In my previous example, if I got from my back end system a localisation with a  "ageOfLocationInfo"= 19, if you asked me with maxAge = 0  I should probably sent an explicit 'unable to localize' instead of  "verificationResult": false. This could need additional discussion because from a boolean we can switch to a string with an enum (true/false/unableToProvide) .

jlurien:
> Hello @jlurien Regarding your question, my understanding is:
> 
> * value 0 mean 'now' (real time localisation)
> * it should be the value default.
> * mandatory or not? I haven't a strong opinion but I understood that https://github.com/patrice-conil has a preference for the mandatory version.
> * as we use 3GPP Monitoring event, the age could be useful when the location type is "LAST_KNOWN_LOCATION" (and not "CURRENT_LOCATION").  Indeed if I got from my back end system a localisation with a  "ageOfLocationInfo"= 19, if you asked me with maxAge = 0  I cannot localize you while if you asked me with maxAge = 20 I send back  "verificationResult": true. I'm not sure about the maximum ageOfLocalisation we are able to manage in the BE.
> * Completely agree with your point about defining also the expected behavior when the requested age cannot be fulfilled. In my previous example, if I got from my back end system a localisation with a  "ageOfLocationInfo"= 19, if you asked me with maxAge = 0  I should probably sent an explicit 'unable to localize' instead of  "verificationResult": false. This could need additional discussion because from a boolean we can switch to a string with an enum (true/false/unableToProvide) .

My concerns here about API usability and "dev friendliness":

If default is 0, meaning "current location" (I prefer this term instead of real-time), and we make maxAge required, we are forcing clients to explicitly provide a maxAge. So if default is the best service, then I wonder in which cases a client would ask for anything worse on purpose. Only maybe if clients get a first response of type "unable to provide" and then ask again for "last know location". Then, a better design approach would probably be to provide the maxAge indication as part of the response, instead of providing it in the request. Clients may not be familiar with typical ages in our networks to decide which threshold in minutes is enough to get the last known location if available.

ukkaps:
@bigludo7 , @jlurien PR is approved now and I do not have write access. can one of you merge this PR now?

jlurien:
@bigludo7, @ukkaps, 

Sorry, but I had some pending comments, and there is still not agreement about how to add this new parameter. I will revert this commit and we can discuss further on this on tomorrow's meeting. For other PRs, please let's wait until we have at least the approval of at least 2 people.

In the meantime, if you could read my latest comments and follow the discussion, it would be great. Thanks

jlurien:
Hi @bigludo7,

- The current proposal considers `maxAge` as an **input** parameter, so the response would depend on the requested threshold about age. But maybe we could think if it fits better as part of the response. Quoting myself :) "Clients may not be familiar with typical ages in our networks to decide which threshold in minutes is enough to get the last known location if available."
- For the second point I guess you assume that network location is fresh, not stalled at A,B. In that case is clearly false, yes.  

The concept of considering the age makes sense but we have to decide how to fit it in the API design, request or response, optional or required, default or not, any associated error if needed, etc. We are not in a hurry for this so let's try to find the best approach.

PR #11, opened 2023-01-10
Create Location_API.md

jlurien:
The agreed version of this API missed the API doc md that we have for other APIs. I followed the usual template.

### DIFF
--- a/documentation/API_documentation/Location_API.md
+++ b/documentation/API_documentation/Location_API.md
@@ -1,0 +1,101 @@
+# Overview
+This API provides the customer with the ability to verify the location of a device. 
+
+## 1\. Introduction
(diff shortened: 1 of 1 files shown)


bigludo7 (commented):
Look almost good. 
Only one point: the swagger features a 404 that you did not documented.

BTW I'm not sure about 404 usage - for a request to a subscription which is not managed by the network provider?

jlurien:
> Look almost good. Only one point: the swagger features a 404 that you did not documented.
> 
> BTW I'm not sure about 404 usage - for a request to a subscription which is not managed by the network provider?

Yes, I missed it according to the yaml. Really I think 404 makes no sense for this endpoint as there are no path parameters which could point to a not found resource. All error stuff is just copied and pasted from other APIs, so was not well thought. Anyway, for this V0.1 I will align doc with yaml and let's work on better errors in future versions.

PR #22, opened 2023-02-07
YAML Update: add UNKNOWN for verify location

VijayKesharwani:
Adding "unknown" https://github.com/camaraproject/DeviceLocation/issues/19
It can happen that the network can't locate the device, because it is offline and the location is also not cached. In this there should be coming the right response from the API.
Changes
Adding to true or false the return value unkown, when it is not possible to get the location


bigludo7 (commented):
Hello
Thanks for this. For me adding unknown value is a good. 2 comments:

- We should use same case in the swagger:  true, false, UNKNOWN (upper case or minor case but same)
- Documentation (.md) should be updated (line) 48 to list the 3 values;

Thanks

jlurien (approved):
LGTM

bigludo7 (approved):
Works for me

Issue #18, opened 2023-01-31
Subscription for Location Area


NoelWirzius:
**Summary**
===========
Create a subscription mode for the DeviceLocation API. So that developers are informed if a device is leaving or entering the "radius". This is important for tracking devices for Presetting of Homesettings or for tracking of logistics and more. 

**Changes** 
==========
For this there are several changes to be done

* Adding new paths `/subscriptions`
  * Brining in new optional parameter
    * Based on `/verify` 
    * `EventType` is an array which is giving the subscription mode 
      * `AREA_ENTERED` = the device is leaving the defined radius 
      * `AREA_LEFT` = the device is entering the defined radius 
    * ExpireTime to set how long the subscription should be active 
  *  Activate Multiple Devices and Multiple Events

bigludo7:
Hello @NoelWirzius 
I'm also interested to contribute.

For the area definition, do you think we can stick to the radius or do we have to extend with polygon (or multi polygon) definition ?

RogerJL:
How many devices should one be able to subscribe too?
Suppose all companies subscribes to all their registered users - ready to push ads when entering...

Note: I think the EventTypes are mixed up in the suggestion.

akoshunyadi:
As decided in https://github.com/camaraproject/DeviceLocation/issues/64 the endpoint should be implemented in a separate API
The name is to be discussed

shilpa-padgaonkar:
please check this and provide feedback https://github.com/camaraproject/Commonalities/issues/8#issuecomment-1629615969

jgarciahospital:
Please considere this message as a formal inclusion of the GSMA OGW Product WS request to include Subscription/notification mechanism as part of OGW services. 

https://github.com/camaraproject/Commonalities/issues/8#issuecomment-1642419272

For the specific case of location verification, the request includes device location verification API to include such subscription/notification mechanism, which is usually called "geofencing" in certain fora. Examples of such funcionality in trems of input-outputs are also included, to be aligned with current subscription mechanism in this API. 

Additional context
More details included in the formal request file, attached:

[Camara Submission - Subscription_Notifications v2.docx](https://github.com/camaraproject/DeviceLocation/files/12098193/Camara.Submission.-.Subscription_Notifications.v2.docx)

ukkaps:
[...]

PR #24, opened 2023-02-28
add /subscription-endpoint for AREA_ENTERED and AREA_LEFT-events

maxl2287:
This PR is related to #18 .
Created an endpoint to subscribe for location-area-events.

### DIFF
--- a/code/API_definitions/location.yaml
+++ b/code/API_definitions/location.yaml
@@ -17,7 +17,7 @@ security:
 #  - BasicAuth: []
 #  - apiKey: []
   - three_legged:
-    - device-location-read
+      - device-location-read
 servers:
   - url: '{apiRoot}/{basePath}'
     variables:
@@ -59,6 +59,53 @@ paths:
           $ref: '#/components/responses/Generic500'
         '503':
           $ref: '#/components/responses/Generic503'
+  /subscriptions:
+    post:
+      tags:
+        - Location Subscription
+      summary: 'Create a location subscription for a device'
+      operationId: createSubscription
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/CreateSubscription'
+        required: true
+      responses:
+        '201':
+          description: Created (Successful creation of subscription)
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/SubscriptionInfo'
+        '400':
+          $ref: '#/components/responses/Generic400'
+        '401':
+          $ref: '#/components/responses/Generic401'
+        '403':
+          $ref: '#/components/responses/Generic403'
+        '409':
+          $ref: '#/components/responses/Generic409'
+        '500':
+          $ref: '#/components/responses/Generic500'
+        '503':
+          $ref: '#/components/responses/Generic503'
+  /notifications:
+    post:
+      tags:
+        - Notification Callback
+      summary: "Subscription notification callback"
+      operationId: postNotification
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/Notification"
+      responses:
+        '204':
+          description: No Content
+
 components:
   securitySchemes:
     oAuth2ClientCredentials:
@@ -66,15 +113,15 @@ components:
       flows:
         clientCredentials:
           tokenUrl: '{tokenUrl}'
-          scopes: {}
-#    BasicAuth:
+          scopes: { }
+    #    BasicAuth:
 #      type: http
-#      scheme: basic
-#    apiKey:
-#      type: apiKey
-#      description: API key to authorize requests
-#      name: apikey
-#      in: query
+    #      scheme: basic
+    #    apiKey:
+    #      type: apiKey
+    #      description: API key to authorize requests
+    #      name: apikey
+    #      in: query
     three_legged:
       type: oauth2
       flows:
@@ -122,6 +169,11 @@ components:
         ipv6Addr:
           $ref: '#/components/schemas/Ipv6Addr'
       minProperties: 1
+    UeIdList:
+      description: List of User equipment identifiers
+      type: array
+      items:
+        $ref: '#/components/schemas/UeId'
     ExternalId:
       type: string
       example: '123456789@domain.com'
@@ -143,9 +195,7 @@ components:
     Ipv6Addr:
       type: string
       format: ipv6
-      allOf:
-        - pattern: '^((:|(0?|([1-9a-f][0-9a-f]{0,3}))):)((0?|([1-9a-f][0-9a-f]{0,3})):){0,6}(:|(0?|([1-9a-f][0-9a-f]{0,3})))(\/(([0-9])|([0-9]{2})|(1[0-1][0-9])|(12[0-8])))?$'
-        - pattern: '^((([^:]+:){7}([^:]+))|((([^:]+:)*[^:]+)?::(([^:]+:)*[^:]+)?))(\/.+)?$'
+      pattern: '(^((:|(0?|([1-9a-f][0-9a-f]{0,3}))):)((0?|([1-9a-f][0-9a-f]{0,3})):){0,6}(:|(0?|([1-9a-f][0-9a-f]{0,3})))(\/(([0-9])|([0-9]{2})|(1[0-1][0-9])|(12[0-8])))?$)|(^((([^:]+:){7}([^:]+))|((([^:]+:)*[^:]+)?::(([^:]+:)*[^:]+)?))(\/.+)?$)'
       example: '2001:db8:85a3:8d3:1319:8a2e:370:7344'
       description: |
         IPv6 address, following IETF 5952 format, may be specified in form <address/mask> as:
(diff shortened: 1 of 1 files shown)


bigludo7 (commented):
Hello
Thanks for this proposal.
Few comments:
- We should probably add a new scope for three_legged (device-location-tracking)
- We could add startedAt attribute in the request (optional) to manage UC when the tracking is not required at the time of the request.
- I did not see in your proposal where you define the area. For me we should have an area attribute (type object) that can be either a [a point + radius] (for circle) or [array of points] (for polygon) or [array of [array of points]] (for array of polygon). I will suggest to use a oneOf with discriminator.
- Last but not least should not we try to adopt a common notification structure for all API (suggested here : https://github.com/camaraproject/WorkingGroups/issues/156) - I'm ok with your attributes name but I imagine something like this:
`
{
  "subscriptionId": "string",

  "eventType": "DEVICE_LOCATION_TRACKING_AREA_ENTERED",

  "eventTime": "2023-02-28T14:49:24.112Z"

  "eventDetail": {
        "ueId": {
             "externalId": "123456789@domain.com",
             "msisdn": "123456789",
            "ipv4Addr": "192.168.0.1/24",
           "ipv6Addr": "2001:db8:85a3:8d3:1319:8a2e:370:7344"
       }
  }
}
`

Looking to discuss this in next meeting.

maxl2287:
Hi @bigludo7,

thanks for your good feedback!
I have some questions about your suggestions:

1.)
> We could add startedAt attribute in the request (optional) to manage UC when the tracking is not required at the time of the request.

What is the effort of adding here a startedAt? What does the staretedAt stands for? The start of the subscription-request?

2.)
> I did not see in your proposal where you define the area. For me we should have an area attribute (type object) that can be either a [a point + radius] (for circle) or [array of points] (for polygon) or [array of [array of points]] (for array of polygon). I will suggest to use a oneOf with discriminator.

I would add here, like for the /verify, a property "area" which will hold the values for "longitude", "latitude" and "accurrency".
Would this also fit for you? Than we have a defined radius.
```json
  "area": {
    "latitude": 50.735851,
    "longitude": 7.10066,
    "accuracy": 50
  }
```
3.)
>  I'm ok with your attributes name but I imagine something like this:
{ "subscriptionId": "string", "eventType": "DEVICE_LOCATION_TRACKING_AREA_ENTERED", "eventTime": "2023-02-28T14:49:24.112Z" "eventDetail": { "ueId": { "externalId": "123456789@domain.com", "msisdn": "123456789", "ipv4Addr": "192.168.0.1/24", "ipv6Addr": "2001:db8:85a3:8d3:1319:8a2e:370:7344" } } }

A short question to the naming of the eventType: 
Why is "AREA_ENTERED" not good enough? "DEVICE_LOCATION_TRACKING" seems to be what this project here is about, so it seems for me senseless to put this information at the beginning of the enum (if you not putting all eventTypes at one place, in a common repository or something like that).

Thanks in advance for an answer! 😺

BR
Max

bigludo7:
Hi @maxl2287 

Thanks for your consideration of my review :)

1) My proposal to add the startedAt is to tackle case where the tracking has not to be performed at request time but later. I imagine some logistic UC to follow container and the tracking should start in 3 days from now. Not a big issue for me if you think is not worthwhile in this first release ;)

2) For me it would be good if we are to define either a circle (like you did) or a polygon

For a circle :
```json
  "area": {
    "latitude": 50.735851,
    "longitude": 7.10066,
    "accuracy": 50
    "type": "Circle"
  }
```
or for a polygon:
```json
  "area": {
    "polygon" : [
          {"latitude": 50.735851,
          "longitude": 7.10066},
          {"latitude": 52.15987,
          "longitude": 7.10066},
          {"latitude": 52.15987,
          "longitude": 7.98066},
          {"latitude": 50.735851,
          "longitude": 7.98066},
         ]
    "accuracy": 50
    "type": "Polygon"
  }
```
It means that in the swagger definition the type attribute is the discriminator and we provide both circle & polygon definition in the swagger.

3) Suggestion to add DEVICE_LOCATION_TRACKING prefix (btw DEVICE_LOCATION is just good enough) is because of the proposal to use a generic event structure for all CAMARA API, so the evenType should be explicit. for me the notification endpoint (on client side) could not specialized to a particular API but be able to receive any CAMARA api notification. Anyway, if you think AREA_ENTERED is self sufficient I can live with it.

code/API_definitions/location.yaml
```diff
@@ -200,6 +250,92 @@ components:
...
+      description: List of Location Event Type
+      type: array
+      items:
+        $ref: '#/components/schemas/LocationEventType'
+      example: [ AREA_ENTERED, AREA_LEFT ]
+    CreateSubscription:
+      description: The request for creating a Location Subscription
+      type: object
+      properties:
+        ueId:
```
jlurien, line 271:
```suggestion
        ueIds:
```
If a list we should use a plural, or ueIdList

code/API_definitions/location.yaml
```diff
@@ -200,6 +250,92 @@ components:
...
+          $ref: '#/components/schemas/UeIdList'
+        eventType:
```
jlurien, line 273:
```suggestion
        eventTypes:
```

or eventTypeList

code/API_definitions/location.yaml
```diff
@@ -200,6 +250,92 @@ components:
...
+      enum:
+        - AREA_ENTERED
+        - AREA_LEFT
+    LocationEventTypeList:
+      description: List of Location Event Type
+      type: array
+      items:
+        $ref: '#/components/schemas/LocationEventType'
+      example: [ AREA_ENTERED, AREA_LEFT ]
+    CreateSubscription:
```
jlurien, line 267: The area is missing

jlurien (commented):
I added some comments as a first iteraration. Some I see that are also indicated by Ludovic. We need also a mechanism to unsubscribe events.

jlurien:
> Hi @maxl2287
> 
> Thanks for your consideration of my review :)
> 
> 1. My proposal to add the startedAt is to tackle case where the tracking has not to be performed at request time but later. I imagine some logistic UC to follow container and the tracking should start in 3 days from now. Not a big issue for me if you think is not worthwhile in this first release ;)
> 2. For me it would be good if we are to define either a circle (like you did) or a polygon
> 
> For a circle :
> 
> ```json
>   "area": {
>     "latitude": 50.735851,
>     "longitude": 7.10066,
>     "accuracy": 50
>     "type": "Circle"
>   }
> ```
> 
> or for a polygon:
> 
> ```json
>   "area": {
>     "polygon" : [
>           {"latitude": 50.735851,
>           "longitude": 7.10066},
>           {"latitude": 52.15987,
>           "longitude": 7.10066},
>           {"latitude": 52.15987,
>           "longitude": 7.98066},
>           {"latitude": 50.735851,
>           "longitude": 7.98066},
>          ]
>     "accuracy": 50
>     "type": "Polygon"
>   }
> ```
> 
> It means that in the swagger definition the type attribute is the discriminator and we provide both circle & polygon definition in the swagger.
> 
> 3. Suggestion to add DEVICE_LOCATION_TRACKING prefix (btw DEVICE_LOCATION is just good enough) is because of the proposal to use a generic event structure for all CAMARA API, so the evenType should be explicit. for me the notification endpoint (on client side) could not specialized to a particular API but be able to receive any CAMARA api notification. Anyway, if you think AREA_ENTERED is self sufficient I can live with it.

We should align the area definitions with the rest os the API. If we support polygons here, they should be supported for the verify endpoints as well.

bigludo7:
> We should align the area definitions with the rest os the API. If we support polygons here, they should be supported for the verify endpoints as well.

Hello @jlurien 
Yes for me it makes sense to align the verify operation.

bigludo7 (commented):
Look very good.
Updated the API Version.
I did not mean to remove existing scope but just add another one ;)
Thanks

maxl2287:
I also updated some wrong examples in /verify-endpoint and updated the plural in the /subscriptions - endpoint

bigludo7:
One quick thing.
Do you think we need to add after line 316:
```yaml
      discriminator:
        propertyName: 'type'
        mapping:
         Circle: '#/components/schemas/Circle'
         Polygon: '#/components/schemas/Polygon'
```

I should confess I do not know if it's better to be explicit or not.
@maxl2287 & @jlurien thoughts?

maxl2287:
Maybe a dumb question, but I do not know yet what the advantage of discriminator is?
Never had experience with that, so maybe you can give me please a short explaination what effort it could bring here. 

https://swagger.io/docs/specification/data-models/inheritance-and-polymorphism/ --> Didn't helped me yet 😄

jlurien:
> Maybe a dumb question, but I do not know yet what the advantage of discriminator is? Never had experience with that, so maybe you can give me please a short explaination what effort it could bring here.
> 
> https://swagger.io/docs/specification/data-models/inheritance-and-polymorphism/ --> Didn't helped me yet 😄

If you define that certain request or response contains an object property with a schema, which is a "base class", such as "Area":
        area:
          $ref: '#/components/schemas/Area'

any specific instance will really be of one of the children classes (Polygon or Circle). The discriminator is the property within the base class that tells the child class of the specific instance, then the instance has to comply with the schema of that child class.

In Swagger 2.0, there was a limitation and the values of the discriminator property had to be exactly the name of the child class (Polygon or Circle, case sensitive). In OAS3 you can define a mapping, and the values can be different from the class name, for example:

      discriminator:
        propertyName: 'type'
        mapping:
         CIRCLE: '#/components/schemas/Circle'
         POLYGON: '#/components/schemas/Polygon'

If you use the same values as in the PR , the mapping is redundant but it is not harmful. I'm not sure if we have guidelines for the format of enum values, in many APIS they are UPPER_CASE but not sure if this is consistent everywhere.

jlurien:
In order to not make this PR so complex I suggest to separate the support of Polygons as Area to a separare issue. Here we can focus on the aspects of the subscription model, and any reference to Area will follow the global model that we define for that.

Wdyt @bigludo7 and @maxl2287 ?

bigludo7:
> In order to not make this PR so complex I suggest to separate the support of Polygons as Area to a separate issue. Here we can focus on the aspects of the subscription model, and any reference to Area will follow the global model that we define for that.
> 
> Wdyt @bigludo7 and @maxl2287 ?

I let both of you decide for this question - I fine to have it both or to separate in 2 PR. The important point for me is to align the documentation once agreed on the swagger.

jlurien:
I prefer to discuss evolution for Area definition as a separate issue. Let's assume here that there is a #/components/schemas/Area schema and we can refer to it anywhere. 

I can open an issue for the Area discussion. I have also some new feedback from our Product people that is part of that discussion.

maxl2287:
@jlurien & @bigludo7 :

Shell I revert this PR to just support without the "area".
Or shell I create a new PR and we let this PR open with the "area" inside?

bigludo7:
Hello
If i understand correctly:
https://github.com/camaraproject/DeviceLocation/pull/27 for the subscription
https://github.com/camaraproject/DeviceLocation/issues/26 for the area

and we close this PR #24 

Works for me ! thanks !

jlurien:
Perfect @maxl2287. Associated with #26 I will create a first PR to add the new Area model, then you can refer to it in the new PR for subscription.

Issue #26, opened 2023-03-01
Evolution of Area definition


jlurien:
Derived from discussion in #24

So far (as of v0.1.0), we have 3 properties to define the area, which is always a circle centered in (`latitude`, `longitude`) with radius `accuracy`:

        latitude:
          $ref: '#/components/schemas/Latitude'
        longitude:
          $ref: '#/components/schemas/Longitude'
        accuracy:
          $ref: '#/components/schemas/Accuracy

In order to support more areas, we can consider Area as a polymorphic object, for example:

```
Area:
  description: The geometrical area considered for device location
  oneOf:
    - $ref: '#/components/schemas/Circle'
    - $ref: '#/components/schemas/Polygon'
    - $ref: '#/components/schemas/PostCode'
  discriminator:
    propertyName: type   
```

Apart from the current Circle, there are proposals to support Polygon (Orange) and PostCode (Telefónica)

bigludo7:
Very strange @maxl2287 - I cannot assign you to this issue !

bigludo7:
I've discussed with people familiar with this topic and they pointed me to : https://geojson.org/ 
Seems good materiel to start with something and not far to what we have.

jlurien:
I can take care of this task @bigludo7. Good to have some standard to look into.

I will make a first PR just adding support for a new object `Area`, which by now it can only be a Circle. This can be then reference in the Subscriptions PR, and we can work later on expanding other sub-types of Area in separate tracks.

jlurien:
Closing issue as new Polymorphic model for Area is already merged

PR #29, opened 2023-03-02
Add support for polymorphic Areas

jlurien:
As introduced in #26, I have prepared a model to evolve the Area concept:

- Initially only the current Circle is considered
- Intention of this PR is to agree on the polymorphic structure to support more types of Area in the near future (Polygon, Postcode...)

Preview of current proposal with Redoc
image

Use of `oneOf` and `allOf` and how they can be combined is a bit confusing in the spec. I opted to create an `Area` object which must be `oneOf` the sub-types, and create a `BaseArea` object to hold the common properties, which is only the `type`.

With Swagger 2.0, where `oneOf` was not available, polymorphism could be expressed only with `allOf`, just referring to the base class. Still, some code generators work better without the use of `oneOf` but I think is more aligned with OAS3 how it is now. We should check if implementations work well with this model.

### DIFF
--- a/code/API_definitions/location.yaml
+++ b/code/API_definitions/location.yaml
@@ -38,7 +38,7 @@ paths:
         content:
           application/json:
             schema:
-              $ref: '#/components/schemas/VerifyLocationRequest'               
+              $ref: '#/components/schemas/VerifyLocationRequest'
         required: true
       responses:
         '200':
--- a/code/API_definitions/location.yaml
+++ b/code/API_definitions/location.yaml
@@ -8,7 +8,7 @@ info:
   license:
     name: Apache 2.0
     url: https://www.apache.org/licenses/LICENSE-2.0.html
-  version: 0.1.0
+  version: 0.2.0
 externalDocs:
   description: Product documentation at Camara
   url: https://github.com/camaraproject/
@@ -38,7 +38,7 @@ paths:
         content:
           application/json:
             schema:
-              $ref: '#/components/schemas/VerifyLocationRequest'
+              $ref: '#/components/schemas/VerifyLocationRequest'               
         required: true
       responses:
         '200':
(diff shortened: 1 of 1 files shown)


jlurien:
> Thanks @jlurien
> 
> I understood we will not follow GeoJson correct? I'm always a bit anxious when we are not aligned to existing standard but I checked on [google map API](https://developers.google.com/maps/documentation/geolocation/overview) and their definition is close to the one you have proposed:
> 
> ```json
> {
>   "location": {
>     "lat": 37.421875199999995,
>     "lng": -122.0851173
>   },
>   "accuracy": 120
> }
> ```
> 
> One comment 'inside the design'
> 
> Your proposed this design:
> 
> ```yaml
>   schemas:
>     Area:
>       oneOf: 
>         -  $ref: "#/components/schemas/Circle"
>       discriminator: 
>         propertyName: type
>     BaseArea:
>       type: object
>       description: Delimited area
>       required:
>         - type
>       properties:
>         type:
>           type: string
>           description:  Type of this area. Supported types are `Circle`
>       discriminator: 
>         propertyName: type
>     Circle:
>       allOf:
>         - $ref: '#/components/schemas/BaseArea'
>         - type: object
>           description: Circular area
>           required:
>             - center
>             - radius
>           properties:
>             center:
>               $ref: '#/components/schemas/Point'
>             radius:
>               type: number
>               description: Accuracy from center expected for location verification in km
>               minimum: 2
>               maximum: 200
> ```
> 
> Are we comfortable with **type** twice defined as discriminator?
> 
> Another design to your consideration:
> 
> ```yaml
>   schemas:
>     Area:
>       oneOf: 
>         -  $ref: "#/components/schemas/Circle"
>       discriminator: 
>         propertyName: type
>       mapping:
>          Circle: "#/components/schemas/Circle"
>     Circle:
>       allOf:
>         - type: object
>           description: Circular area
>           required:
>             - center
>             - radius
>             - type
>           properties:
>             center:
>               $ref: '#/components/schemas/Point'
>             radius:
>               type: number
>               description: Accuracy from center expected for location verification in km
>               minimum: 2
>               maximum: 200
>             type:
>               type: string
> ```
Thanks!

I didn't see a definition for circle in GeoJSON, is there one?. For polygons we may follow their concept of circular line where the first and last points are the same.  I didn't know the GMaps model also, but I think we could adopt it. Indeed I thought initially in the names `lat` and `long`,  but as we are being verbose lately with properties naming, I chose the full ones. But for Polygons, if there is a long array of points, maybe having shorter key names is clearer: [{lat: 1.1223, long: 2.2.22}, {lat: 1.2233, long: 2.234} ...]

The model that you propose (removing `allOf` as it not necessary anymore) may be enough in our scenarios as only `type` is defined in the base class, so we are not saving too much reusing that schema. I will check that preview is still working fine.

bigludo7 (approved):
Thanks @jlurien

PR #31, opened 2023-03-03
Cover partial matches in verification result

jlurien:
As discussed in https://github.com/camaraproject/DeviceLocation/issues/20

- New enum value with an optional new property
- Examples and descriptions enhanced


bigludo7 (commented):
A very very small request... change 'must' to 'could' for matchRate if we're in PARTIAL answer. In order to avoid to be to normative on ourselves.

PR #33, opened 2023-03-09
Add multiple device in verify location

chia2017:
This PR is related to this issue: https://github.com/camaraproject/DeviceLocation/issues/32

### DIFF
--- a/code/API_definitions/location.yaml
+++ b/code/API_definitions/location.yaml
@@ -46,7 +46,16 @@ paths:
           content:
             application/json:
               schema:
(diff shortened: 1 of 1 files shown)


bigludo7 (commented):
Thanks for this contribution.
For me the points highlighted by @jlurien here https://github.com/camaraproject/DeviceLocation/issues/32 should be cleared.

chia2017:
We decided to close this request because we want to go on with another approach regarding multiple device in verify location.

Issue #39, opened 2023-03-16
Add Postal Code as a type of Area


jlurien:
Postal codes are a common user friendly denominations for known areas. 

Proposal is to add a new type of Area `PostalCode`

```
PostalCode:
  type: object
  required:
    - type
    - postalCode
  properties:
    type:
      type: string
      enum:
        - PostalCode
      description: Type of this area
    postalCode:
      type: string
      description: Postal Code (aka Zip code or Postcode), according to the local format. 
    countryCode:
      type: string
      pattern: '^[A-Z]{2}$' 
      description: ISO 3166-1 alpha-2 code for the country. Uppercase. If not specified, each implementation may apply a default value.
```

Optionally a country code may be needed in some global scenarios. Otherwise, each implementation may assume its own country.
There is no global pattern for postal codes. We could define some limits minLength , maxLength, apparently longest one is 10, but it may be better to leave all format verification to implementations.

bigludo7:
To provide my feedback.... in France we have postalCode but another code is used for geographic purpose: the INSEE code (the postal code is more an administrative one).
Perhaps same situation could occur in other country. 
We have several options to accommodate this:

1) make a model more generic with for example a `code` attribute (to provide code value) and a `codeType`  (to provide the ype of code provided). 

````yaml
GeoCode:
  type: object
  required:
    - type
    - code
    - codeType
  properties:
    type:
      type: string
      enum:
        - GeoCode
      description: Type of this area
    code:
      type: string
      description: Code (aka Zip code or Postcode), according to the code type. 
    codeType:
      type: string
      description: Type of code (could be Postcode, INSEE Code, etc...), 
    countryCode:
      type: string
      pattern: '^[A-Z]{2}$' 
      description: ISO 3166-1 alpha-2 code for the country. Uppercase. If not specified, each implementation may apply a default value.
````

2/ leverage the polymorphic model and provide explicit specialization. In this case we keep your proposal and for example I add a structure for inseeCode.

I think that for a developer, option 2 is probably better as more explicit, but we need to find a way to expose by code the model managed by an implementation (If Telefonica managed Circle & postalCode while Orange manage Circle & insee code and then DT Circle & polygon ...developer should be able to get this information on the fly).

JoachimDahlgren:
There is no global standard for postal codes and it does not always point to a geographical area (and in cases it does point to a geographical area the size thereof can vary a lot depending on if it is a rural or urban area). 

[...]

PR #47, opened 2023-04-05
Add new area type: PostalCode

jlurien:
Proposal for new Area type: Postal Code, as discussed in https://github.com/camaraproject/DeviceLocation/issues/39

### DIFF
--- a/code/API_definitions/location.yaml
+++ b/code/API_definitions/location.yaml
@@ -108,14 +108,17 @@ components:
   schemas:
     Area:
       oneOf: 
-        -  $ref: "#/components/schemas/Circle"
+        - $ref: "#/components/schemas/Circle"
+        - $ref: "#/components/schemas/PostalCode"
       discriminator: 
         propertyName: type
         mapping: 
           Circle:  "#/components/schemas/Circle"
+          PostalCode: "#/components/schemas/PostalCode"
+      description: Portion of Earth surface delimited by certain borders. Different type of areas may be supported with different geometries. `Circle` is always supoported. 
     Circle:
       type: object
-      description: Circular area
+      description: Circular area specified by a center (`location`) and a radius (`accuracy`).
       required:
         - type
         - location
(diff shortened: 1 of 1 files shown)


bigludo7 (approved):
Good for me with few typo fixes provided

code/API_definitions/location.yaml
```diff
@@ -164,7 +169,30 @@ components:
...
+          enum:
+            - PostalCode
+          description:  Type of this area.
+        postalCode:
+          type: string
+          description: Postal Code (aka Zip code or Postcode), according to the local format.
+        countryCode:
+          type: string
+          pattern: "^[A-Z]{2}$"
+          description: ISO 3166-1 alpha-2 code for the country. Uppercase. If not specified, each implementation may assume a default value.
```
akoshunyadi, line 191: 2 issues with countryCode:
- in the DeviceStatus API it means the MCC code and countryName means the ISO 3166 code
- I think it would be better to make this parameter mandatory, otherwise the API call without this parameter will have different results for the different provider implementations

jlurien: It is good to align terminology among APIs. To me this is more a "code" than a "name", but if the other term is agreed and consolidated in the Commonalities Glossary I'm open to align it.
I'm also open to make the property required.

RogerJL:
Would vote against this if I could, see discussion https://github.com/camaraproject/DeviceLocation/issues/39#issue-1627238412

jlurien:
> Would vote against this if I could, see discussion [#39 (comment)](https://github.com/camaraproject/DeviceLocation/issues/39#issue-1627238412)

We have not established any voting process for issues but any feedback is appreciated. We are discussing issues and PRs regularly on our bi-weekly calls. Which company do you represent in CAMARA?

I will keep the PR on hold by now, until we have better clarification on how to support many type of areas and their coexistence.

jlurien:
New PR will be created after discussion in #83 with new requirements based on GSMA request.

PR #48, opened 2023-04-05
Create MoM-2023-04-11.md

jlurien:


bigludo7 (commented):
Good for me but will approve after the meeting.
For INSEE this is the code used here but did not get 'push' from product management to implement it for now so I will not propose a specialization.
For notification pattern we're progressing and we have an ad hoc call on this topic April 12th.

I will comment this in next meeting.

jlurien:
@bigludo7 Updated with your minutes. Ready to be merged

PR #55, opened 2023-05-09
Swagger definition for Get device location

bigludo7:
First stab for API definition for **retrieving a device location**. There are points to be commented in this PR and discussed in team meeting:

- I've not yet provided any documentation - will do once we agreed on the interface directly in the OAS.
- Use of device identifiers as defined in latest QoD version (phoneNumber an not MSISDN) - this introduce a gap with the check api.
- Definition of a specific scope distinct from the check operation
- In the response only lat/log & accuracy are provided. I opted to provided straightforward information and not use area schema (with oneOf) --> could be challenged.
- We need some update on the other api - for example the little for the existing api should be check device location API instead of device location API. 
- etc.

### DIFF
--- a/code/API_definitions/Get DeviceLocation.yaml
+++ b/code/API_definitions/Get DeviceLocation.yaml
@@ -100,7 +100,8 @@ components:
           $ref: '#/components/schemas/Device'
         maxAge:
           type: integer
-          description: Maximum age of the location information which is accepted for the verification (in seconds)
+          description: Maximum age of the location information which is accepted for the location retrieval (in seconds)
+
           minimum: 60
       required:
         - device
--- a/code/API_definitions/Get DeviceLocation.yaml
+++ b/code/API_definitions/Get DeviceLocation.yaml
@@ -26,7 +26,7 @@ servers:
         default: location/v0
         description: Base path for the device location API
 paths:
-  /retrieve-location:
+  /retrieve:
     post:
       tags:
         - Location retrieval
--- a/code/API_definitions/Get DeviceLocation.yaml
+++ b/code/API_definitions/Get DeviceLocation.yaml
@@ -1,6 +1,6 @@
 openapi: 3.0.3
 info:
-  title: Get Device location API
+  title: Retrieve Device location API
   description: Service Enabling Network Function API for location retrieval from a device identifier.
   termsOfService: http://swagger.io/terms/
   contact:
@@ -28,59 +28,18 @@ servers:
         default: location/v0
         description: Base path for the device location API
 paths:
-  /locations:
-    get:
+  /location:
+    post:
       tags:
         - Location retrieval
-      summary: 'Retrieve location for a user equipment'
+      summary: 'Execute location retrieval for a user equipment'
       operationId: retrieveLocation
-      description: Retrieve location for an user equipement from an identifier like phone number, IP address (and port) or external identifier provided by the network operator.
-      parameters:
-          - description: Subscriber number in E.164 format (starting with country code). Optionally prefixed with '+'.
-            name: phoneNumber
-            in : query
-            schema: 
-              type: string
-              pattern: '^\+?[0-9]{5,15}$'
-          - description: Identifier assigned by the mobile network operator (MNO) for the device 
-            name: networkAccessIdentifier
-            in: query
+      requestBody:
+        content:
+          application/json:
             schema:
-              type: string
-          - description:  |
-              IPv6 address, following IETF 5952 format, may be specified in form <address/mask> as:
-                - address - The /128 subnet is optional for single addresses:
-                  - 2001:db8:85a3:8d3:1319:8a2e:370:7344
-                  - 2001:db8:85a3:8d3:1319:8a2e:370:7344/128
-                - address/mask - an IP v6 number with a mask:
-                  - 2001:db8:85a3:8d3::0/64
-                  - 2001:db8:85a3:8d3::/64
-            name: ipv6Address
-            in : query
-            schema:
-              type: string
-              pattern: 
-                -'^((:|(0?|([1-9a-f][0-9a-f]{0,3}))):)((0?|([1-9a-f][0-9a-f]{0,3})):){0,6}(:|(0?|([1-9a-f][0-9a-f]{0,3})))(\/(([0-9])|([0-9]{2})|(1[0-1][0-9])|(12[0-8])))?$'
-                -'^((([^:]+:){7}([^:]+))|((([^:]+:)*[^:]+)?::(([^:]+:)*[^:]+)?))(\/.+)?$'
-          - description: |
-              IPv4 address may be specified in form <address/mask> as:
-              - address - an IPv4 number in dotted-quad form 1.2.3.4. Only this exact IP number will match the flow control rule.
-              - address/mask - an IP number as above with a mask width of the form 1.2.3.4/24.
-              In this case, all IP numbers from 1.2.3.0 to 1.2.3.255 will match. The bit width MUST be valid for the IP version.
-            name: Ipv4Address
-            in : query
-            schema:
-              type: string
-              pattern: 
-                -'^((:|(0?|([1-9a-f][0-9a-f]{0,3}))):)((0?|([1-9a-f][0-9a-f]{0,3})):){0,6}(:|(0?|([1-9a-f][0-9a-f]{0,3})))(\/(([0-9])|([0-9]{2})|(1[0-1][0-9])|(12[0-8])))?$'
-                -'^((([^:]+:){7}([^:]+))|((([^:]+:)*[^:]+)?::(([^:]+:)*[^:]+)?))(\/.+)?$'
-          - description: User equipment port. Device port may be required along with IP address to identify the target device
-            name: port
-            in : query
-            schema:
-              type: integer
-              minimum: 0
-              maximum: 65535
+              $ref: '#/components/schemas/RetrievalLocationRequest'
+        required: true
       responses:
         '200':
           description: Location retrieval result
@@ -96,7 +55,6 @@ paths:
                     latitude: 50.735851
                     longitude: 7.10066
                     accuracy: 2
-
         '400':
           $ref: '#/components/responses/Generic400'
         '401':
@@ -134,6 +92,67 @@ components:
           scopes:
             device-location-provide: Provide device location
   schemas:
+    RetrievalLocationRequest:
+      type: object
+      properties:
+        ueId:
+          $ref: '#/components/schemas/UeId'
+        uePort:
+          $ref: '#/components/schemas/Port'
+      required:
+        - ueId
+    UeId:
+      description: User equipment identifier
+      type: object
+      properties:
+        externalId:
+          $ref: '#/components/schemas/ExternalId'
+        msisdn:
+          $ref: '#/components/schemas/MSISDN'
+        ipv4Addr:
+          $ref: '#/components/schemas/Ipv4Addr'
+        ipv6Addr:
+          $ref: '#/components/schemas/Ipv6Addr'
+      minProperties: 1
+    ExternalId:
+      type: string
+      example: '123456789@domain.com'
+    MSISDN:
+      type: string
+      pattern: '^\+?[0-9]{5,15}$'
+      example: '123456789'
+      description: Subscriber number in E.164 format (starting with country code). Optionally prefixed with '+'.
+    Ipv4Addr:
+      type: string
+      format: ipv4
+      pattern: '^([01]?\d\d?|2[0-4]\d|25[0-5])(?:\.(?:[01]?\d\d?|2[0-4]\d|25[0-5])){3}(\/([0-9]|[1-2][0-9]|3[0-2]))?$'
+      example: '192.168.0.1/24'
+      description: |
+        IPv4 address may be specified in form <address/mask> as:
+          - address - an IPv4 number in dotted-quad form 1.2.3.4. Only this exact IP number will match the flow control rule.
+          - address/mask - an IP number as above with a mask width of the form 1.2.3.4/24.
+            In this case, all IP numbers from 1.2.3.0 to 1.2.3.255 will match. The bit width MUST be valid for the IP version.
+    Ipv6Addr:
+      type: string
+      format: ipv6
+      allOf:
+        - pattern: '^((:|(0?|([1-9a-f][0-9a-f]{0,3}))):)((0?|([1-9a-f][0-9a-f]{0,3})):){0,6}(:|(0?|([1-9a-f][0-9a-f]{0,3})))(\/(([0-9])|([0-9]{2})|(1[0-1][0-9])|(12[0-8])))?$'
+        - pattern: '^((([^:]+:){7}([^:]+))|((([^:]+:)*[^:]+)?::(([^:]+:)*[^:]+)?))(\/.+)?$'
+      example: '2001:db8:85a3:8d3:1319:8a2e:370:7344'
+      description: |
+        IPv6 address, following IETF 5952 format, may be specified in form <address/mask> as:
+          - address - The /128 subnet is optional for single addresses:
+            - 2001:db8:85a3:8d3:1319:8a2e:370:7344
+            - 2001:db8:85a3:8d3:1319:8a2e:370:7344/128
+          - address/mask - an IP v6 number with a mask:
+            - 2001:db8:85a3:8d3::0/64
+            - 2001:db8:85a3:8d3::/64
+    Port:
+      type: integer
+      minimum: 0
+      maximum: 65535
+      description: User equipment port. Device port may be required along with IP address to identify the target device
+      example: 5060
     RetrievalLocationResponse:
       type: object
       required:
--- a/code/API_definitions/Get DeviceLocation.yaml
+++ b/code/API_definitions/Get DeviceLocation.yaml
@@ -291,7 +291,7 @@ components:
                 message: 'Invalid input'
             MaxAgeIssue:
               value:
-                tatus: 400
+                status: 400
                 code: LOCATION_RETRIEVAL.MAXAGE_INVALID_ARGUMENT
                 message: 'maxAge threshold cannot be satisfied'
     Generic401:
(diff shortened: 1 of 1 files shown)


code/API_definitions/Get DeviceLocation.yaml
```diff
@@ -0,0 +1,237 @@
...
+        description: Base path for the device location API
+paths:
+  /locations:
+    get:
+      tags:
+        - Location retrieval
+      summary: 'Retrieve location for a user equipment'
+      operationId: retrieveLocation
+      description: Retrieve location for an user equipement from an identifier like phone number, IP address (and port) or external identifier provided by the network operator.
+      parameters:
```
jlurien, line 38: Here we have to decide whether we model this as GET or as POST. GET is more natural options to retrieve data but it has the problem of modelling complex input parameters. If we wanted to reuse the same `device` object that we have in QoD or the original spec of Location, we could pass an object as query parameter, which is valid, but it has to be encoded properly, which is less readable. Reusing the object as it is has the advantages that it is possible to express the constraints between the different properties, for example that port is required along with public IP address.
Modelling as POST is also possible and align with other APIs but guidelines are not very clear about this. We could model the endpoint as POST /retrieve-location, so options would be:

UrlEncoded( GET location/v0/locations?device={"ipV4Address":{"public_ip_address":"1.2.3.4","public_port":1234}})

or

POST /location/v0/retrieve-location
{
  "device": {
    "ipV4Address": {
      "public_ip_address": "1.2.3.4",
      "public_port": 1234
    }
  }
)

There are similar cases in other APIs so maybe it makes sense to raise this in commonalities and adjust guidelines accordingly

code/API_definitions/Get DeviceLocation.yaml
```diff
@@ -0,0 +1,237 @@
...
+          scopes:
+            device-location-provide: Provide device location
+  schemas:
+    RetrievalLocationResponse:
+      type: object
+      required:
+        - latitude
+        - longitude
+        - accuracy
+      properties:
```
jlurien, line 143: This is the most obvious response. Not sure if in some cases a different Area other than Circle would make sense, if network can return a different geometry.

jlurien (commented):
Thanks for providing first approach. My main doubt is how to model data retrieval with complex inputs, GET vs POST, a recurrent debate while designing APIs :)

code/API_definitions/Get DeviceLocation.yaml
```diff
@@ -0,0 +1,237 @@
...
+  - three_legged:
+    - device-location-provide
+servers:
+  - url: '{apiRoot}/{basePath}'
+    variables:
+      apiRoot:
+        default: http://localhost:9091
+        description: API root
+      basePath:
+        default: location/v0
```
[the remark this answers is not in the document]
bigludo7: Yes agreed - When I crafted this proposal I've also feeling that we need to update the check location API accordingly.
Pragmatically do you think, @jlurien , that I should do this in same PR ?

I tend to think yes but looking for guidance.

code/API_definitions/Get DeviceLocation.yaml
```diff
@@ -0,0 +1,237 @@
...
+        description: Base path for the device location API
+paths:
+  /locations:
```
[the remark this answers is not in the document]
bigludo7: Yes agrees - note this is related to previous point as existing check location is a POST` /localisations `with a s. (but only one device checked). i have keep consistency.
My preference is also to have both endpoints in singular.

code/API_definitions/Get DeviceLocation.yaml
bigludo7: Same for me. I guess circle is well enough for first release. 

Let's wait for UC requiring other Area definition and in this case we'll implement the `oneOf `model.

code/API_definitions/Get DeviceLocation.yaml
bigludo7: Yes this is a point to be discussed. 
Looking in other API we have POST in device status right now and GET for Device-identifier

i will raise this point in commonalities but not sure we'll get a rule.

Regarding our case, I will probably prefer to use POST pattern to align with Check. 

Let's tab this point for next meeting.

bigludo7:
Discussion in Commonalities --> https://github.com/camaraproject/WorkingGroups/issues/212

bigludo7 (commented):
Update after first set of review.
Mainly changed GET operation to POST

bigludo7 (commented):
Renamed resource retrieve-location (instead of location) as using POST operation

code/API_definitions/Get DeviceLocation.yaml
JoachimDahlgren: Agree that POST is probably a better way as the input is complex and it aligns with both device status as well as with verify location.

code/API_definitions/Get DeviceLocation.yaml
JoachimDahlgren: Point is probably good as a starting point. There is however a potential complexity in providing only a point. Depending on the southbound system a request for location might come in another format, such as a polygon,  which then  needs to be translated into a point with an appropriate accuracy.

code/API_definitions/Get DeviceLocation.yaml
```diff
@@ -0,0 +1,256 @@
...
+      flows:
+        authorizationCode:
+          authorizationUrl: https://auth.example.com/authorize
+          tokenUrl: https://auth.example.com/token
+          scopes:
+            device-location-provide: Provide device location
+  schemas:
+    RetrievalLocationRequest:
+      type: object
+      properties:
```
JoachimDahlgren, line 97: Should the UeId be renamed to Device to be aligned with QoD 0.9?

code/API_definitions/Get DeviceLocation.yaml
```diff
@@ -0,0 +1,256 @@
...
+        ueId:
+          $ref: '#/components/schemas/UeId'
+        uePort:
+          $ref: '#/components/schemas/Port'
+      required:
+        - ueId
```
JoachimDahlgren, line 103: UeId should probably be aligned with QoD 0.9. Be renamed to Device with properties phoneNumber, networkAccessIdentifier, ipv4Address, ipv6Address with Port moving to the ipv4Address as publicPort.

bigludo7: This is a fair comment. I've aligned with our existing check location API but probably we should align both device location api with Qod v0.9 if this later is enough mature.

bigludo7 (commented):
Alignement with QoD device identification structure.

bigludo7 (commented):
Alignement with QoD device identification structure.

bigludo7:
@JoachimDahlgren @jlurien @rartych

- changed operation to POST
- aligned with QoD device identifier structure.

I suggest to tackle check-device device identifier structure alignment in another PR once this one merged.
Then I will create 2 other PR to add subscription management.

code/API_definitions/Get DeviceLocation.yaml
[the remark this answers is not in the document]
jlurien: For decisions which affect both APIs I think it is ok to make changes in same PR

code/API_definitions/Get DeviceLocation.yaml
```diff
@@ -0,0 +1,292 @@
...
+      
+        In all cases, publicAddress must be specified, along with at least one of either privateAddress or publicPort, dependent upon which is known. In general, mobile devices cannot be identified by their public IPv4 address alone.
+
+    SingleIpv4Addr:
+      type: string
+      description: A single IPv4 address with no subnet mask
+      example: "84.125.93.10"
+      pattern: '^(([0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])$'
+
+    Ipv4Address:
```
jlurien, line 157: This schema is not used

bigludo7 (commented):
Fixed issues identified by @jlurien

code/API_definitions/Get DeviceLocation.yaml
bigludo7: Yes correct _ fixed

code/API_definitions/Get DeviceLocation.yaml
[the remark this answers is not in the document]
jlurien: Thought better of it, let's make separate PRs for changes in  the other API. There is already one to adapt ueId to device.

code/API_definitions/Get DeviceLocation.yaml
```diff
@@ -0,0 +1,280 @@
...
+  - url: '{apiRoot}/{basePath}'
+    variables:
+      apiRoot:
+        default: http://localhost:9091
+        description: API root
+      basePath:
+        default: location/v0
+        description: Base path for the device location API
+paths:
+  /retrieve-location:
```
akoshunyadi, line 31: I have to note here, that our guidelines says: "Verbs use is not allowed."
Maybe: location-detection or location-retrieval

code/API_definitions/Get DeviceLocation.yaml
akoshunyadi: Shouldn't we add a time info of this location data? I think this will be needed sooner or later, since it can be quite important for the clients.

code/API_definitions/Get DeviceLocation.yaml
akoshunyadi: @jlurien ok, then it's fine for me ( in the location API there was also a 'verify' initially )

code/API_definitions/Get DeviceLocation.yaml
bigludo7: @akoshunyadi good question - For me it depends on if we want the API to provides current location or not last known location (associated to an age info).  This is tricky because depending on the business case one or the other could be required. 
I was thinking to keep thing simple and just provide current location or 404 if not able to provide but open to discuss this (add for example an `ageOfLocation `attribute in the response)

@jlurien WDYT?

bigludo7 (commented):
Applied decisions done by the team June 21th:
Change accuracy unit (from km to meters)
Add maxAge in request (in seconds - min 60)
Add ageOfLocation in response

bigludo7:
Hello @jlurien @akoshunyadi @JoachimDahlgren @rartych 
Kind reminder
I've applied all required changes 2 weeks ago. 
Looking for your review as we planned to release this API in July
Thanks

akoshunyadi (approved):

code/API_definitions/Get DeviceLocation.yaml
```diff
@@ -0,0 +1,290 @@
...
+          type: number
+          format: double
+          minimum: -180
+          maximum: 180
+          example: 7.10066     
+        accuracy:
+          type: number
+          description: Expected accuracy for the location in meters, from the point defined by latitude/longitude
+        lastLocationTime: 
+          description: last date and time when the device was localized (using ISO 8601 with timezone)
```
jlurien, line 211:
```suggestion
          description: Last date and time when the device was localized (using ISO 8601 with timezone)
```

code/API_definitions/Get DeviceLocation.yaml
```diff
@@ -0,0 +1,290 @@
...
+        longitude:
+          description: Longitude component of location
+          type: number
+          format: double
+          minimum: -180
+          maximum: 180
+          example: 7.10066     
+        accuracy:
+          type: number
+          description: Expected accuracy for the location in meters, from the point defined by latitude/longitude
```
jlurien, line 209:
```suggestion
          description: Provided accuracy for the location, in meters, from the point defined by latitude/longitude

```

code/API_definitions/Get DeviceLocation.yaml
```diff
@@ -0,0 +1,290 @@
...
+            device-location-provide: Provide device location
+  schemas:
+    RetrievalLocationRequest:
+      type: object
+      properties:
+        device:
+          $ref: '#/components/schemas/Device'
+        maxAge:
+          type: integer
+          description: Maximum age of the location information which is accepted for the verification (in seconds)
```
jlurien, line 103:
```suggestion
          description: Maximum age of the location information which is accepted for the location retrieval (in seconds)

```

code/API_definitions/Get DeviceLocation.yaml
```diff
@@ -0,0 +1,290 @@
...
+        status:
+          type: integer
+          description: HTTP status code returned along with this error response
+        code:
+          type: string
+          description: Code given to this error
+        message:
+          type: string
+          description: Detailed error description
+  responses:
```
jlurien, line 230: We should define an error when the `maxAge` threshold cannot be satisfied.

jlurien (commented):
Some suggestions

code/API_definitions/Get DeviceLocation.yaml
```diff
@@ -0,0 +1,280 @@
+openapi: 3.0.3
+info:
+  title: Retrieve Device location API
+  description: Service Enabling Network Function API for location retrieval from a device identifier.
```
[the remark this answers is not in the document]
bigludo7: Yes - will do once we validate the OAS.

bigludo7 (commented):
Following review:
- Changed scope name
- Added specific error for maxAge
- Integrated inheritance mode for retrieved area

bigludo7:
@jlurien : updated the OAS - looking for your feedback in particular for the integration of the inheritance model for the area.

bigludo7 (commented):
fixed example typo

bigludo7:
@jlurien Everything look resolved from my perspective - Ok for you? Waiting for your approval to merge it

jlurien:
> @jlurien Everything look resolved from my perspective - Ok for you? Waiting for your approval to merge it

Let's discuss it tomorrow. Merging it now may impact the releasing process for v0.2. We may prefer to wait until v0.2 for location-verification is released and then merge the first PRs for the new APIs in the family.

code/API_definitions/Get DeviceLocation.yaml
```diff
@@ -0,0 +1,346 @@
...
+#      name: apikey
+#      in: query
+    three_legged:
+      type: oauth2
+      flows:
+        authorizationCode:
+          authorizationUrl: https://auth.example.com/authorize
+          tokenUrl: https://auth.example.com/token
+          scopes:
+            device-location-provide: Provide device location
```
jlurien, line 97:
```suggestion
            location-retrieval-read: Provide device location
```

jlurien (approved):
Mostly OK. I have some comments to align "aesthetically" both location-verification and this API, in terms of title, path name (verb vs plural name), scope (it is ok in security but not in securitySchemes)

bigludo7 (commented):
Updated following @jlurien review.

code/API_definitions/Get DeviceLocation.yaml
```diff
@@ -0,0 +1,346 @@
...
+          $ref: '#/components/responses/Generic503'
+components:
+  securitySchemes:
+    oAuth2ClientCredentials:
+      type: oauth2
+      flows:
+        clientCredentials:
+          tokenUrl: '{tokenUrl}'
+          scopes: {}
+#    BasicAuth:
```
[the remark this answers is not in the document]
bigludo7: Done !

jlurien (dismissed):
LGTM

code/API_definitions/Get DeviceLocation.yaml
```diff
@@ -0,0 +1,344 @@
...
+  - url: '{apiRoot}/{basePath}'
+    variables:
+      apiRoot:
+        default: http://localhost:9091
+        description: API root
+      basePath:
+        default: location/v0
+        description: Base path for the device location API
+paths:
+  /retrieve-location:
```
jlurien, line 29:
```suggestion
  /retrieve:
```
Agreed after #84, to consolidate naming in both APIs

jlurien (commented):
Rename path to just /retrive, as location-verification has been renamed to /verify

bigludo7 (commented):
Changed resource name to /retrieve

bigludo7 (commented):
Alignment for oAuth2AuthorizationCode & scope as done in PR #88

code/API_definitions/Get DeviceLocation.yaml
```diff
@@ -0,0 +1,344 @@
...
+    name: Apache 2.0
+    url: https://www.apache.org/licenses/LICENSE-2.0.html
+  version: 0.1.0-wip
+externalDocs:
+  description: Product documentation at Camara
+  url: https://github.com/camaraproject/
+security:
+  - oAuth2ClientCredentials: []
+  - oAuth2AuthorizationCode:
+    - location-retrieval:read
```
jlurien, line 18: If the API name is just "location", then scope would be "location:retrieve"

code/API_definitions/Get DeviceLocation.yaml
```diff
@@ -0,0 +1,344 @@
...
+#      name: apikey
+#      in: query
+    oAuth2AuthorizationCode:
+      type: oauth2
+      flows:
+        authorizationCode:
+          authorizationUrl: https://auth.example.com/authorize
+          tokenUrl: https://auth.example.com/token
+          scopes:
+            location-retrieval:read: Provide device location
```
jlurien, line 95: location:retrieve

code/API_definitions/Get DeviceLocation.yaml
```diff
@@ -0,0 +1,344 @@
...
+          examples:
+            InvalidArgument:
+              value:
+                status: 400
+                code: INVALID_ARGUMENT
+                message: 'Invalid input'
+            MaxAgeIssue:
+              value:
+                status: 400
+                code: LOCATION_RETRIEVAL.MAXAGE_INVALID_ARGUMENT
```
jlurien, line 293: If we follow also API_NAME.{CODE}  then it would be just LOCATION.MAXAGE_INVALID_ARGUMENT

jlurien (commented):
All comments around API name, depending on whether it is just "location" or is "location-retrieval". It affects scopes and error codes naming. If we choose "location-retrieval" then servers must be updated.

bigludo7 (commented):
changer server name to `location-retrieval`

jlurien (approved):
I think it is ready to merge as v0.1-wip

bigludo7:
@jlurien Agreed.
I will wait till Monday and then proceed (and work on the documentation part)
As it is still in wip we can still accommodate comment after - if any.

PR #57, opened 2023-05-22
Create MoM-2023-05-23.md

jlurien:
To be updated after meeting


hdamker:
@jlurien You can create the PR as draft (or convert it to draft) instead using "don't merge" while this is only the agenda.

PR #67, opened 2023-06-19
Doc/mom 2023 06 20

jlurien:
#### What type of PR is this?

* documentation

Agenda and minutes for regular meetings


jlurien:
@bigludo7 Minutes updated with your comments, only one comment:

- Do you recall if we agreed on moving units for accuracy to meters? I felt so but it was mixed with discussion about minimum limit and no sure about outcome.  In the minutes I added that it was decided but I can change it if we are in doubt

bigludo7 (approved):
Thanks
Yes we decided to move to meters.

PR #71, opened 2023-07-03
Doc: Embed documentation into API spec

jlurien:
#### What type of PR is this?

* documentation

- Ad doc to `info.description` section
- Remove legacy doc markdown file

Comply with Commonalities guidelines.

Review that wording is correct. Suggest clarifications or improvements

#### Changelog input

```
API doc is now embedded into API spec 

```


bigludo7 (approved):
With a proposal change to add bullet points.

alpaycetin74 (commented):
May I ask 2 questions:

- Do we not intend to define how the match rate should be calculated? Our definition seems open to interpretation to the implementer.   
- Why is the match partial if the network can provide "higher" accuracy ? Do you mean the numerical value in the response is larger ? This is actually "worse" accuracy than requested. This may be misunderstood.
Thank you.

jlurien:
> May I ask 2 questions:
> 
> * Do we not intend to define how the match rate should be calculated? Our definition seems open to interpretation to the implementer.
> * Why is the match partial if the network can provide "higher" accuracy ? Do you mean the numerical value in the response is larger ? This is actually "worse" accuracy than requested. This may be misunderstood.
>   Thank you.

We should try to agree on a common understanding and implementation. Indeed, there is an open discussion on how to approach the situation with worse accuracy by the network (higher value here is really the opposite). We may open a dedicate discussion for implementation guidelines, and for the time being I'll rephrase the current description to make it more neutral.

bigludo7 (approved):
LGFM

PR #77, opened 2023-07-04
fix/ location-verification: remodel-Area-inheritance

### LINKED ISSUE #72: Remodel Area as inheritance
**Problem description**

Area uses a polymorphic model (oneOf)

**Expected behavior**

Guidelines favor inheritance (allOf) 

**Alternative solution**

Remodel Area a base class and Circle as child

**Additional context**

jlurien:
#### What type of PR is this?

* correction

* Remodel Area as inheritance model with allOf

Fixes #72 

type moved to base class

#### Changelog input

n/a

#### Additional documentation 

n/a


bigludo7 (commented):
Did same work for the raw location and seems were aligned - good if you can take a look to make sure about this alignment.

3 minor requests for your consideration:
- Instead of `type `I renamed the attribute as `areaType`
- Instead of `location `(line 126) could we use `center `as we describe a circle.
- For same reason instead of `accuracy  `could we rename the attribute `radius`.

This more explicit for the API consumer.

jlurien:
> Did same work for the raw location and seems were aligned - good if you can take a look to make sure about this alignment.
> 
> 3 minor requests for your consideration:
> 
> * Instead of `type `I renamed the attribute as `areaType`

OK, indeed I was already considering this name, just in case we have for example some `codeType` or similar in the future.
 
> * Instead of `location `(line 126) could we use `center `as we describe a circle.
> * For same reason instead of `accuracy  `could we rename the attribute `radius`.
> 

Modelling areas as geometrical figures, it could make sense. Originally it was meant to represent an specific target point with some accuracy, which could be understood as the device more likely to be close to the center. If we add other type of areas, as code areas, is just another area.

I will ask internally for other opinions.

> This more explicit for the API consumer.

jlurien:
@bigludo7 I have renamed the Area and Circle properties with your suggestions

bigludo7 (approved):
Look good for me

bigludo7 (approved):
LGTM

PR #74, opened 2023-07-04
New API for location subscriptions

chia2017:
#### What type of PR is this?
enhancement/feature

#### What this PR does / why we need it:
Create a new API for creating subscriptions for location based events

#### Which issue(s) this PR fixes:
This PR is related to issue #18  Subscription for Location Area 

Fixes #

#### Changelog input

```
 release-note

```

#### Additional documentation 

This section can be blank.

```
docs

```

PR #46, opened 2023-03-31
Update location.yaml

bigludo7:
Aligned code from the error structure in the example with the value provided in the guideline.

Fix issue: #45

PR #88, opened 2023-08-16
Update location-verification.yaml

jlurien:
#### What type of PR is this?

* correction

Adjust security scopes naming to api_name:[resource:]action, to be aligned with suggestions for #74 

n/a

If OK, we should also adjust #55

### DIFF
--- a/code/API_definitions/location-verification.yaml
+++ b/code/API_definitions/location-verification.yaml
@@ -51,8 +51,8 @@ externalDocs:
   url: https://github.com/camaraproject/
 security:
   - oAuth2ClientCredentials: []
-  - three_legged:
-    - location-verification-read
+  - oAuth2AuthorizationCode:
+    - location-verification:verify
 servers:
   - url: '{apiRoot}/{basePath}'
     variables:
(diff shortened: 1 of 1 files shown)


bigludo7 (approved):
Hello @jlurien 
I like the idea to standardize the security scope across all APIs but one comment: Is it only a DeviceLocation project decision or a global CAMARA one ? probably this is something that we should propose to Commonalities?

jlurien:
> Hello @jlurien I like the idea to standardize the security scope across all APIs but one comment: Is it only a DeviceLocation project decision or a global CAMARA one ? probably this is something that we should propose to Commonalities?

I think it is something pending to agree on consistent guidelines for security, securitySchemes and scopes across all WGs. There is already an [issue](https://github.com/camaraproject/Commonalities/issues/46) on Commonalities for the scope format. One thing to review also is when to allow clientCredentials or authorizationCode. There is a debate about this in Identity WG.

To me it would be more flexible to define openidConnect as securityScheme and define scopes for it. In the end, OIDC is probably what we will use and it relies on a .well_known endpoint lo learn about the supported authentication flows

PR #92, opened 2023-08-30
Add documentation + update date time attribute description

### LINKED ISSUE #40: Provide raw location in geographical coordinates
Originating from the brief discussion in the meeting 2023-03-14.

In addition to being able to verify the location, Ericsson propose that a developer should also be able to get the raw location information of a device as known by the network.

This would serve use cases in which the device does not have built in GNSS support. Extended with subscription support it would also serve use cases for tracking such devices.

Location data should be provided in form of geographical coordinates (x, y, z). The reason for this is that network identities such as CellID etc. are typically not known by developers (and should perhaps not even be exposed externally). The location could be given as a point or a shape with uncertainty area.

Location is in most jurisdictions considered to be sensitive data and thereby consent by device owner/user must be verified before providing it to the developer.

Further more we believe that the API to access location data should be aligned with API to get device status information to make it easy for developers to request these at the same time (originally we have proposed them as one and the same API https://github.com/camaraproject/DeviceStatus/issues/17

Location could be reported as for instance (but not limited to)

{
       "ageOfLocationInfo": 0,
             "geographicArea": {
             "shape": "POINT",
             "point": {
                     "lon": -180,
                     "lat": -90
              }
      }
}

or 

{
       "ageOfLocationInfo": 10,
             "geographicArea": {
             "shape": "POINT_UNCERTAINTY_CIRCLE",
             "point": {
                     "lon": -180,
                     "lat": -90
              },
             "uncertainty": 200
      }
}

bigludo7:
#### What type of PR is this?

Added Documentation + Added precision about use of RFC3339 for the date time.

Added Documentation + Added precision about use of RFC3339 for the date time.

Fixes #40 

#### Changelog input

```
- added documentation for the API
- precise use of RFC 3339 for the datetime attribute

```

#### Additional documentation 

This section can be blank.

```
docs

```

### DIFF
--- a/code/API_definitions/location-retrieval.yaml
+++ b/code/API_definitions/location-retrieval.yaml
@@ -1,7 +1,51 @@
 openapi: 3.0.3
 info:
-  title: Device location retrieval API
-  description: Service Enabling Network Function API for location retrieval from a device identifier.
+  title: Location retrieval API
+  description: |
+    This API provides the ability to retrieve a device location. 
+
+    # Introduction
+
+    With this API, customers can retrieve the area where a certain user device is localized. The area provided in the response is described by a circle determined by coordinates (latitude and longitude) and an accuracy defined by the radius. 
+    
+    The requester could optionally ask for a freshness of the localization information by providing a maxAge ("I want a location not older than 600 seconds")
+
+    The result accuracy depends on the network's ability and accuracy to locate the device. 
+    
+    Additionally to location information, the answer will also provide indication about the location time. 
+
+    Location retrieval API could be useful in scenarios such as:
+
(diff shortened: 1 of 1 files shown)


jlurien (approved):
LGTM, sorry for the delay

bigludo7:
@jlurien Thanks - no problem

@rartych @JoachimDahlgren @alpaycetin74 : Do you plan to make a review ?

alpaycetin74:
> @jlurien Thanks - no problem
> 
> @rartych @JoachimDahlgren @alpaycetin74 : Do you plan to make a review ?

Apologies, this was at the beginning of my vacation and I missed your message.

PR #115, opened 2023-11-14
move "org.camaraproject.geofencing.v0.subscription-ends" out of /subscriptions - request

### LINKED ISSUE #113: Geofencing API - Subscription-ends should not a subscribe-able event type
**Problem description**
The `org.camaraproject.geofencing.v0.subscription-ends` notification type did not require a subscription. Indeed this notification type should be always send when the subscription terminates. As such this type must not be listed in the POST /Subscriptions (but this type could be of course used in the notification)

**Expected behavior**
In the yaml have a separate array for the event type in the subscription and the one for the notification in order to not allow `org.camaraproject.geofencing.v0.subscription-ends`  in the `POST /souscriptions`

**Alternative solution**

**Additional context**

maxl2287:
#### What type of PR is this?

* correction

- Removes the event "org.camaraproject.geofencing.v0.subscription-ends" out of the _/subscriptions_ - request. 
Only use this event for CloudEvent - callbacks.
- Replace single-quotes with double-quotes

Fixes #113

### DIFF
--- a/code/API_definitions/geofencing.yaml
+++ b/code/API_definitions/geofencing.yaml
@@ -69,7 +69,7 @@ externalDocs:
   url: https://github.com/camaraproject/
 
 servers:
-  - url: '{apiRoot}/geofencing/v0'
+  - url: "{apiRoot}/geofencing/v0"
     variables:
       apiRoot:
         default: http://localhost:9091
@@ -83,17 +83,17 @@ paths:
     post:
       tags:
         - Geofencing subscriptions
-      summary: 'Create a geofencing subscription for a device'
+      summary: "Create a geofencing subscription for a device"
       operationId: createSubscription
       requestBody:
         content:
           application/json:
             schema:
-              $ref: '#/components/schemas/CreateSubscription'
+              $ref: "#/components/schemas/CreateSubscription"
         required: true
       callbacks:
         event-notifications:
-          '{$request.body#/webhook/notificationUrl}':
+          "{$request.body#/webhook/notificationUrl}":
             post:
               summary: "Subscription notification callback"
               description: |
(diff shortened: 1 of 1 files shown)


bigludo7 (approved):
LGTM 
Thanks @maxl2287

code/API_definitions/geofencing.yaml
```diff
@@ -441,26 +441,33 @@ components:
...
         area-left - Event triggered when the device leaves the given area
-        subscription-ends - Event triggered when the subscription ends
-      type: string
       enum:
         - org.camaraproject.geofencing.v0.area-entered
         - org.camaraproject.geofencing.v0.area-left
-        - org.camaraproject.geofencing.v0.subscription-ends
+
+    SubscriptionEventType:
+      allOf:
```
jlurien, line 464: We cannot "extend" an enum this way, as allOf requires compliance to both schemas, and any of the specific values cannot comply with both schemas at the same time. We have to define a new enum with the 3 values, same model as in https://github.com/camaraproject/DeviceStatus/blob/main/code/API_definitions/device-status.yaml#L568

jlurien (changes requested):

code/API_definitions/geofencing.yaml
```diff
@@ -441,26 +441,38 @@ components:
...
+        area-entered - Event triggered when the device enters the given area
+        
+        area-left - Event triggered when the device leaves the given area
+        
+        subscription-ends - Event triggered when the subscription ends
       enum:
         - org.camaraproject.geofencing.v0.area-entered
         - org.camaraproject.geofencing.v0.area-left
-        - org.camaraproject.geofencing.v0.subscription-ends
+        - org.camaraproject.device-status.v0.subscription-ends
```
jlurien, line 475:
```suggestion
        - org.camaraproject.geofencing.v0.subscription-ends
```

jlurien (changes requested):

jlurien (approved):
LGTM

bigludo7 (approved):
LGTM

PR #128, opened 2023-12-19
Update README.md

jlurien:
#### What type of PR is this?

* documentation

Adapted for Release Candidate branch

After that I will create a relese-0.2.0-rc branch, so the links will work


bigludo7 (approved):
I did not first read the point about the links ;)  ...so good for me with the new branch to be created.

jlurien:
> I did not first read the point about the links ;) ...so good for me with the new branch to be created.

I saw that other subprojects were creating these -rc branches. Not my initial intention, but it may be useful in case that RC period extends and we want to add future content to the main branch

PR #130, opened 2023-12-19
Update README.md

jlurien:
#### What type of PR is this?

* correction

Remove /blob from links


jlurien:
@bigludo7 when you can, please

bigludo7 (approved):

Issue #141, opened 2024-01-17
Geofencing - Adding `SUBSCRIPTION_DELETED` as a Termination Reason


bigludo7:
**Problem description**
Feedback from implementation.
For the `terminationReason` we have following value: 

- `NETWORK_TERMINATED` - API server stopped sending notification 
- `SUBSCRIPTION_EXPIRED` - Subscription expire time (required by the requester) has been reached. 

We did not have a specific value when the termination was triggered from a DELETE operation.

We can assume that because the consume side perform the DELETE no need for notification but for sake of consistency send an event seems to be a good option.

**Possible evolution**
Add value `SUBSCRIPTION_DELETED` in the `terminationReason` array

**Alternative solution**

**Additional context**

jlurien:
This may have sense, but it is something that would apply to subscriptions for all APIs, wdyt @PedroDiez ?

alpaycetin74:
If we assume deletion of subscription is a synchronous operation, successful response to DELETE operation is sufficient.
If we believe the deletion of subscription in the backend must be done asynchronously, a notification is necessary because successful response to DELETE  would only mean the backend received & accepted the request.

To be flexible, we can support both. But then we should define 2 different success responses to DEL so that the client would now if deletion is completed or they should wait for a notification. 

Having said that, we could also make creation of subscription asynchronous (HTTP 201 created followed by a notification later).
But we never considered it. Maybe it is also necessary for DEL ?

bigludo7:
Hello @alpaycetin74 
For me the question is not about the fact that the DELETE is performed async or sync. As we send a 204 the deletion is sync for the system triggering the DELETE.

But we cannot assume technical correlation between the system listening to the event channel (the target of the notificationUrl) and the system performing the DELETE (OK this is the problem of the consumer). 
But as we are committed to send a suscription-ends notification to the listener, the point for me is to provide information about subscription termination reason to this listener.

Does it make sense?

alpaycetin74:
hello @bigludo7 , 
yes, especially when thinking of  _"_we cannot assume technical correlation between the system listening to the event channel (the target of the notificationUrl) and the system performing the DELETE"_  it makes sense. It could help the listener application to shut down, maybe save some resources/cost/battery.

PedroDiez:
> This may have sense, but it is something that would apply to subscriptions for all APIs, wdyt @PedroDiez ?

[...]

PR #144, opened 2024-01-29
Meeting minutes: MoM-2024-01-30.md

jlurien:
#### What type of PR is this?

* documentation

Agenda for next meeting. to be updated with minutes

### DIFF
--- a/documentation/MeetingMinutes/MoM-2024-01-30.md
+++ b/documentation/MeetingMinutes/MoM-2024-01-30.md
@@ -1,0 +1,87 @@
+# CAMARA DeviceLocation meeting - 2024-01-30
+
+*January 30th, 2024*
+
+## Attendees
+
+TBC
+
+Device Location minutes: [https://github.com/camaraproject/DeviceLocation/tree/main/documentation/MeetingMinutes](https://github.com/camaraproject/DeviceLocation/tree/main/documentation/MeetingMinutes)
+
+## Agenda
+
+* Open issues and PRs
+* Any other business
+  
+## Open issues and PRs
+
+### Review v0.2-rc2
+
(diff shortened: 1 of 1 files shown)


jlurien:
@bigludo7 Please review the meeting minutes. If you need some clarification as you were not in this one, let me know. Thanks

bigludo7 (approved):
@jlurien for https://github.com/camaraproject/DeviceLocation/issues/141 not sure why waiting for Commonalities outputs. 
The guidelines stated: _For this specific event, the data must feature terminationReason attribute._

There is not specification of the value enumeration and with the diversity of the APIs not sure it makes sense . For me  I'm afraid that going to Commonalities will take unnecessary delay for a very short point - WDYT?

jlurien:
> @jlurien for #141 not sure why waiting for Commonalities outputs. The guidelines stated: _For this specific event, the data must feature terminationReason attribute._
> 
> There is not specification of the value enumeration and with the diversity of the APIs not sure it makes sense . For me I'm afraid that going to Commonalities will take unnecessary delay for a very short point - WDYT?

The question is about whether we should agree on the value for `terminationReason` across APIs, as I think that this case will happen for all APIs with subscriptions.

In QoD there is [PR](https://github.com/camaraproject/QualityOnDemand/pull/258) proposal that is somehow similar, to add a new reason for termination, even if  in QoD there are no explicit subscriptions and notifications are implicit to the QoS session, and there the reason value for this case is 

        * `DELETE_REQUESTED`- User requested the deletion of the session before the requested duration expired

They also have NETWORK_TERMINATED as another reason. 

For Geofencing I see SUBSCRIPTION_DELETED acceptable as well, but other APIs with subscriptions may decide on another value if it is not documented in Commonalities.

PR #149, opened 2024-02-07
fix: correct $ref for SubscriptionEventType in geofencing

maxl2287:
#### What type of PR is this?

* bug

Fixes the reference to the component SubscriptionEventType

### DIFF
--- a/code/API_definitions/geofencing.yaml
+++ b/code/API_definitions/geofencing.yaml
@@ -63,7 +63,7 @@ info:
   license:
     name: Apache 2.0
     url: https://www.apache.org/licenses/LICENSE-2.0.html
-  version: 0.2.0-wip
+  version: 0.1.0-rc3
 externalDocs:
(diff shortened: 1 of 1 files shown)


code/API_definitions/geofencing.yaml
```diff
@@ -63,7 +63,7 @@ info:
   license:
     name: Apache 2.0
     url: https://www.apache.org/licenses/LICENSE-2.0.html
-  version: 0.1.0-rc2
+  version: 0.2.0-wip
```
jlurien, line 66:
```suggestion
  version: 0.1.0-wip
```

We have to fix this for 0.1.0 as well so let's keep the current version. We'll have to create an rc3 probably

jlurien (changes requested):

jlurien (approved):

bigludo7 (approved):

PR #153, opened 2024-02-14
Update README.md

jlurien:
#### What type of PR is this?

* correction

Update links as now /blob is no longer part of the raw file URL


jlurien:
@bigludo7 please check if you can approve this, to create rc3. -Thanks

bigludo7 (approved):
LGTM

PR #159, opened 2024-03-04
fix: update the PARTIAL case description that `match_rate` is set in the ewaoibaw

maxl2287:
…e included in the response

#### What type of PR is this?

* bug
* correction
* enhancement/feature
* cleanup
* documentation
* subproject management
* tests

Fixes #

#### Changelog input

```
 release-note

```

#### Additional documentation 

This section can be blank.

```
docs

```

### DIFF
--- a/code/API_definitions/location-verification.yaml
+++ b/code/API_definitions/location-verification.yaml
@@ -12,7 +12,7 @@ info:
     
     * If the network's estimation of the device's location is fully contained within the requested area, the verification result is `TRUE`.
     * If the network's estimation of the device's location does not overlap with the requested area at all, the verification result is `FALSE` . 
-    * If the network's estimation of the device's location partially overlaps with the requested area, or it fully contains the requested area (because it is larger), the result is 'PARTIAL'. In this case, a `match_rate` may be included in the response, indicating an estimation of the likelihood of the match in percent.
+    * If the network's estimation of the device's location partially overlaps with the requested area, or it fully contains the requested area (because it is larger), the result is 'PARTIAL'. In this case, a `match_rate` is included in the response, indicating an estimation of the likelihood of the match in percent.
     * Lastly, the network may not be able to locate the device. In this case, the verification result is `UNKNOWN`.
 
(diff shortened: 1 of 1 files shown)


code/API_definitions/location-verification.yaml
```diff
@@ -12,7 +12,7 @@ info:
     
     * If the network's estimation of the device's location is fully contained within the requested area, the verification result is `TRUE`.
     * If the network's estimation of the device's location does not overlap with the requested area at all, the verification result is `FALSE` . 
-    * If the network's estimation of the device's location partially overlaps with the requested area, or it fully contains the requested area (because it is larger), the result is 'PARTIAL'. In this case, a `match_rate` may be included in the response, indicating an estimation of the likelihood of the match in percent.
+    * If the network's estimation of the device's location partially overlaps with the requested area, or it fully contains the requested area (because it is larger), the result is 'PARTIAL'. In this case, a `match_rate` is included in the response, indicating an estimation of the likelihood of the match in percent.
```
alpaycetin74, line 15: By saying "may be included", we allow the implementer _not to include_ the matchRate even if the match is partial.
If we remove "may be", we make the matchRate _effectively_ mandatory if match is partial.
Is that what we want to do ?

maxl2287: What would be the advantage of having just "PARTIAL" without a match_rate?

maxl2287:
> Good for me but I'm not sure about version update. @jlurien WDYT?
@bigludo7 
What do you suggest?
0.3.0-wip?

code/API_definitions/location-verification.yaml
alpaycetin74: Not really an advantage, just a convenience for the implementer not to implement the calculation part.

jlurien:
Hi, I don't fully remember how intentional was this when we introduced  `matchRate`, as we hadn't fully discussed the implementation details. But I agree that makes more sense to return it when PARTIAL is returned. Regarding versioning it should be for 0.3.0-wip, as 0.2.0 is approved and about to be released.

alpaycetin74 (approved):
LGTM, just let us all be aligned that we make matchRate effectively mandatory in case of partial match. Thank you.

maxl2287:
Version is updated to 0.3.0-wip

bigludo7 (approved):
Thanks @maxl2287

jlurien (approved):
LGTM

PR #166, opened 2024-03-07
Introducing corrections for yamllint

rartych:
#### What type of PR is this?

Add one of the following kinds:
* correction

#### What this PR does / why we need it:
While preparing PR for linting rules, it occurred that `yamllint` produces  large number of errors (mainly trailing spaces).
I propose to merge this changes before introducing the linting rules workflows configuration, otherwise the errors will be raised by Megalinter.

#### Which issue(s) this PR fixes:

Fixes #

#### Changelog input

```
yaml linting errors cleaned up

```

#### Additional documentation 

```
code/API_definitions/geofencing.yaml
  Error: 6:1 [trailing-spaces] trailing spaces
  Error: 8:1 [trailing-spaces] trailing spaces
  Error: 10:1 [trailing-spaces] trailing spaces
  Error: 11:152 [trailing-spaces] trailing spaces
  Error: 12:1 [trailing-spaces] trailing spaces
  Error: 14:1 [trailing-spaces] trailing spaces
  Error: 16:1 [trailing-spaces] trailing spaces
  Error: 18:1 [trailing-spaces] trailing spaces
  Error: 19:55 [trailing-spaces] trailing spaces
  Error: 21:1 [trailing-spaces] trailing spaces
  Error: 23:1 [trailing-spaces] trailing spaces
  Error: 25:1 [trailing-spaces] trailing spaces
  Error: 27:1 [trailing-spaces] trailing spaces
  Error: 29:1 [trailing-spaces] trailing spaces
  Error: 31:1 [trailing-spaces] trailing spaces
  Error: 33:92 [trailing-spaces] trailing spaces
  Error: 34:1 [trailing-spaces] trailing spaces
  Error: 37:114 [trailing-spaces] trailing spaces
  Error: 38:1 [trailing-spaces] trailing spaces
  Error: 41:1 [trailing-spaces] trailing spaces
  Error: 43:1 [trailing-spaces] trailing spaces
  Error: 44:148 [trailing-spaces] trailing spaces
  Error: 45:63 [trailing-spaces] trailing spaces
  Error: 46:161 [trailing-spaces] trailing spaces
  Error: 54:104 [trailing-spaces] trailing spaces
  Error: 55:1 [trailing-spaces] trailing spaces
  Error: 57:1 [trailing-spaces] trailing spaces
  Error: 58:65 [trailing-spaces] trailing spaces
  Error: 464:1 [trailing-spaces] trailing spaces
  Error: 474:1 [trailing-spaces] trailing spaces
[...]

PR #171, opened 2024-03-12
Update location-verification with x-correlator

bigludo7:
#### What type of PR is this?

* enhancement/feature

Update location-verification with x-correlator
This PR partially solve Issue https://github.com/camaraproject/DeviceLocation/issues/160
(2 other PR will come for the 2 other APIs)

Fixes #

Alignement with CAMARA Guidelines

#### Changelog input

```
 release-note
- add x-correlator as parameter for location-verification

```

#### Additional documentation 

This section can be blank.

```
docs

```

### DIFF
--- a/code/API_definitions/location-verification.yaml
+++ b/code/API_definitions/location-verification.yaml
@@ -71,6 +71,8 @@ paths:
       description: |
         Verify whether the location of a device is within a requested area. The operation returns a verification result and, optionally, a match rate estimation for the location verification in percent.
       operationId: verifyLocation
+      parameters:
+        - $ref: '#/components/parameters/x-correlator'
       requestBody:
         required: true
         content:
@@ -108,6 +110,9 @@ paths:
       responses:
         "200":
           description: Location verification result
+          headers:
+            x-correlator:
+              $ref: '#/components/headers/x-correlator'
           content:
             application/json:
               schema:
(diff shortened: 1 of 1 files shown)


jlurien (approved):
LGTM comparing with other APIs, but I opened an issue in Commonalities to discuss about the UUID pattern in the Guidelines https://github.com/camaraproject/Commonalities/issues/166

alpaycetin74:
hello @jlurien @bigludo7  , is this header considered optional ?
It may be useful to make it mandatory in responses, so that the API always returns it to the client for end-to-end troubleshooting.

We have a similar header in our proprietary VF APIs: It is optional in the request, but the API implementation always creates one, prints it in all the logs, and returns it in the API response.
If the API client provides the header in the request, our APIs just use it in logging and mirror it in the response.

bigludo7:
Hello @alpaycetin74 
We're following here what is specified in the commonalities guidelines here: https://github.com/camaraproject/Commonalities/blob/main/documentation/API-design-guidelines.md#9-architecture-headers
and probably it will be good for you:

> When the API Consumer includes the "X-Correlator" header in the request, the API provider must include it in the response with the same UUID as was used in the request. Otherwise, it is optional to include the "X-Correlator" header in the response with any valid (preferably randomly chosen) UUID.

PR #180, opened 2024-04-08
 Geofencing feature file for Test Definition

mdomale:
#### What type of PR is this?

* bug
* correction
* enhancement/feature - Test Definition for Geofencing Feature file
* cleanup
* documentation
* subproject management
* tests

Fixes #

#### Changelog input

```
 release-note

```

#### Additional documentation 

This section can be blank.

```
docs

```

### DIFF
--- a/code/Test_definitions/Geofencing.feature
+++ b/code/Test_definitions/Geofencing.feature
@@ -1,0 +1,118 @@
+#-
+# ---license-start
+# CAMARA Project
+# ---
+# Copyright (C) 2023 - 2024 Contributors | Deutsche Telekom AG to CAMARA a Series
+#
+# The contributor of this file confirms his sign-off for the
+# Developer Certificate of Origin
+#             (http://developercertificate.org).
+# ---
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ---license-end
+
+ @Geofencing 
+Feature: Operations to manage event subscription on geofencing events for leaving and entering in an area
+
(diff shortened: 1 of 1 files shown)


mdomale:
@akoshunyadi  @shilpa-padgaonkar  @hdamker FYI

mdomale:
duplicate hence closing

code/Test_definitions/Geofencing.feature
```diff
@@ -0,0 +1,118 @@
...
+ @Geofencing 
+Feature: Operations to manage event subscription on geofencing events for leaving and entering in an area
+
+  @geofencing_01_Create_geofencing_subscription_for_a_device
+  Scenario:  Create geofencing subscription
+    Given they use the geofencing url
+    When they create subscription
+    Then Response code is 201
+
+  @geofencing_02_Operation_to_retrieve_list_of_subscriptions.--Not yet implemented
```
alpaycetin74, line 33: I guess a space is needed before the double dash, at the moment the string "--Not" is perceived as part of the expression starting with '@'

code/Test_definitions/Geofencing.feature
```diff
@@ -0,0 +1,118 @@
...
+    When they delete subscription for subscription-id
+    Then Response code is 204
+
+  @geofencing_05_Create_invalid_geofencing_subscription_for_a_device
+  Scenario:  Create geofencing subscription with invalid parameter
+    Given they use the geofencing url
+    When they create subscription with invalid parameter
+    Then Response code is 400
+
+  @geofencing_06_Get_invalid_geofencing_subscription_for_a_device
```
alpaycetin74, line 58: We need to clarify what "invalid" subscription id means.
If the id has a valid format, but unknown to the system, the proper response is 404.
If the id does not have a valid format, the proper response is 400.
Maybe we should have one test for each case.

code/Test_definitions/Geofencing.feature
```diff
@@ -0,0 +1,118 @@
...
+  Scenario:  Get geofencing subscription with invalid parameter
+    Given they use the geofencing url
+    When they get subscription with invalid subscription-id
+    Then Response code is 404
+
+
+  @geofencing_07_Delete_invalid_geofencing_subscription_for_a_device
```
alpaycetin74, line 65: Same as in GET case:
We need to clarify what "invalid" subscription id means.
If the id has a valid format, but unknown to the system, the proper response is 404.
If the id does not have a valid format, the proper response is 400.
Maybe we should have one test for each case.

PR #182, opened 2024-04-08
Remove API from title and align use of Device as prefix

### LINKED ISSUE #169: Remove API from info.title
**Problem description**
Currently our APIS include "API" in info.title, which is redundant and not happening with other CAMARA APIs

**Expected behavior**
Keep just the title without API

jlurien:
#### What type of PR is this?

* correction

Align info.title values in the 3 APIs:
- Remove API as suffix
- Add Device as prefix for location-retrieval which was the only one without it
- Use Upper Case for every word, as this is used by most of CAMARA APIs

Fixes #169

### DIFF
--- a/code/API_definitions/geofencing.yaml
+++ b/code/API_definitions/geofencing.yaml
@@ -1,6 +1,6 @@
 openapi: 3.0.3
 info:
-  title: Device geofencing API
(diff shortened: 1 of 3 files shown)


bigludo7 (approved):
LGTM

PR #186, opened 2024-04-11
Bad OAS spec

jlurien:
#### What type of PR is this?

* tests

- Intentionally wrong to test new linting actions


jlurien:
@rartych I merged your PR with the new linting functions and created this PR to test it, but they still don't run, do we need more setup? Thanks

hdamker:
@jlurien thanks ... that answers also my question if the version will be checked (it won't, which is ok, as it allows to use "wip" then).

PR #188, opened 2024-04-23
Adjust maxAge minimum & documentation

### LINKED ISSUE #184: Semantics of the absence of 'maxAge'
**Problem description**
Regarding the usage of 'maxAge' parameter, we have use cases where "real-time" location is needed instead of an x-secs old location. Since 'maxAge' has a minimum value of 60 secs, using 'maxAge' is not a good way for a client to ask for such "real time" location verification.
Our proposal is that not including the 'maxAge' parameter in the request (as it is optional) means that the client is requesting a verification of the "real time" location. I saw the same type of proposal in [this issue](https://github.com/camaraproject/DeviceLocation/issues/42#issuecomment-1480783843), but there were no more explicit mentions to it in the [maxAge issue](https://github.com/camaraproject/DeviceLocation/issues/34). So, ¿is this the original meaning of not including 'maxAge' in the request?.
In the case of real-time verification is requested in this way and it is not available, an 'UNKNOWN' response would be returned and a 'lastLocationTime' could be added if available.
For us it is very important to be able to meet this kind of real-time requirement and, doing in this way, both requirements (real-time without 'maxAge' and lastKnownLocation using 'maxAge') could be met.

**Expected action**
Alignment in the semantics of the spec definition.

bigludo7:
#### What type of PR is this?

* enhancement/feature

Fixes #184 

#### Changelog input

```
 release-note
- Remove minimum 60 seconds to maxAge.
  -  Absence of maxAge means "any age"
  -  maxAge=0 means a fresh calculation.

```

#### Additional documentation 

This section can be blank.

```
docs

```

### DIFF
--- a/code/API_definitions/location-retrieval.yaml
+++ b/code/API_definitions/location-retrieval.yaml
@@ -38,7 +38,7 @@ info:
     * **Max Age**: Maximum age of the location information which is accepted for the location retrieval (in seconds).
 
       * Absence of maxAge means "any age" is acceptable for the client. In other words, this is like maxAge=infinite. In this case the system may still return lastLocationTime, if available. If the system is not able to provide location a error 404 with code LOCATION_RETRIEVAL.DEVICE_NOT_FOUND is send back.
-      * maxAge=0 means a fresh calculation is requested by the client. If the system is not able to provide fresh location (current time - location time > maxDate) an error 422 with code LOCATION_RETRIEVAL.UNABLE_TO_FULFILL_MAX_AGE is send back.
+      * maxAge=0 means a fresh calculation is requested by the client. If the system is not able to provide fresh location an error 422 with code LOCATION_RETRIEVAL.UNABLE_TO_FULFILL_MAX_AGE is send back.
 
     * **Last Location Time** : Last date and time when the device was localized.
 
--- a/code/API_definitions/location-retrieval.yaml
+++ b/code/API_definitions/location-retrieval.yaml
@@ -37,7 +37,7 @@ info:
 
     * **Max Age**: Maximum age of the location information which is accepted for the location retrieval (in seconds).
 
-      * Absence of maxAge means "any age" is acceptable for the client. In other words, this is like maxAge=infinite
+      * Absence of maxAge means "any age" is acceptable for the client. In other words, this is like maxAge=infinite. In this case the system may still return lastLocationTime, if available.
       * maxAge=0 means a fresh calculation is requested by the client.
       If the system does not support it, or fresh location cannot be calculated at that time for any reason, the API response will be "unknown".
 
--- a/code/API_definitions/location-retrieval.yaml
+++ b/code/API_definitions/location-retrieval.yaml
@@ -13,7 +13,7 @@ info:
 
     The retrieved shape depends on the network conditions at the subscriber's location and any of the supported shapes could be received.
 
-    The requester could optionally ask for a freshness of the localization information by providing a maxAge ("I want a location not older than 600 seconds")
+    The requester could optionally ask for a freshness of the localization information by providing a maxAge ("I want a location not older than 600 seconds").
 
     The result accuracy depends on the network's ability and accuracy to locate the device.
 
@@ -36,6 +36,10 @@ info:
     * **Area**: It specifies the geographical surface where a device may be physically located.
 
     * **Max Age**: Maximum age of the location information which is accepted for the location retrieval (in seconds).
+
+      * Absence of maxAge means "any age" is acceptable for the client. In other words, this is like maxAge=infinite
+      * maxAge=0 means a fresh calculation is requested by the client. 
+      If the system does not support it, or fresh location cannot be calculated at that time for any reason, the API response will be "unknown". 
 
     * **Last Location Time** : Last date and time when the device was localized.
 
@@ -146,8 +150,7 @@ components:
           $ref: '#/components/schemas/Device'
         maxAge:
           type: integer
-          description: Maximum age of the location information which is accepted for the location retrieval (in seconds)
-          minimum: 60
+          description: Maximum age of the location information which is accepted for the location retrieval (in seconds). Absence of maxAge means "any age" and maxAge=0 means a fresh calculation.
       required:
         - device
 
--- a/code/API_definitions/location-retrieval.yaml
+++ b/code/API_definitions/location-retrieval.yaml
@@ -37,7 +37,7 @@ info:
 
     * **Max Age**: Maximum age of the location information which is accepted for the location retrieval (in seconds).
 
-      * Absence of maxAge means "any age" is acceptable for the client. In other words, this is like maxAge=infinite. In this case the system may still return lastLocationTime, if available. If the system is not able to provide location a error 400 with code LOCATION_RETRIEVAL.DEVICE_NOT_FOUND is send back. 
+      * Absence of maxAge means "any age" is acceptable for the client. In other words, this is like maxAge=infinite. In this case the system may still return lastLocationTime, if available. If the system is not able to provide location a error 400 with code LOCATION_RETRIEVAL.DEVICE_NOT_FOUND is send back.
       * maxAge=0 means a fresh calculation is requested by the client. If the system is not able to provide fresh location (current time - location time > maxDate) an error 422 with code LOCATION_RETRIEVAL.UNABLE_TO_FULFILL_MAX_AGE is send back.
 
     * **Last Location Time** : Last date and time when the device was localized.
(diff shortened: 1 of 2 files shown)


code/API_definitions/location-retrieval.yaml
```diff
@@ -37,6 +37,10 @@ info:
 
     * **Max Age**: Maximum age of the location information which is accepted for the location retrieval (in seconds).
 
+      * Absence of maxAge means "any age" is acceptable for the client. In other words, this is like maxAge=infinite
+      * maxAge=0 means a fresh calculation is requested by the client.
+      If the system does not support it, or fresh location cannot be calculated at that time for any reason, the API response will be "unknown".
```
alpaycetin74, line 42: I think this behavior can be helpful to add:
"In this case. the system may still return lastLocationTime, if available."

code/API_definitions/location-verification.yaml
```diff
@@ -32,6 +32,12 @@ info:
 
     * **Area**: It specifies the geographical surface where a device may be physically located.
 
+     * **Max Age**: Maximum age of the location information which is accepted for the location verification (in seconds).
+        * Absence of maxAge means "any age" is acceptable for the client. In other words, this is like maxAge=infinite
+        * maxAge=0 means a fresh calculation is requested by the client.
+          If the system does not support it, or fresh location cannot be checked at that time for any reason, the API response will be "unknown".
```
alpaycetin74, line 38: I think this behavior can be helpful to add:
"In this case. the system may still return lastLocationTime, if available."

jlurien: We already indicate that above, with "the verification result will be `UNKNOWN` and the `lastLocationTime` attribute may indicate the last available time for the device location", but it is not bad to reinforce the guideline here

code/API_definitions/location-retrieval.yaml
jlurien, line 42:
```suggestion
      If the system does not support it, or fresh location cannot be calculated at that time for any reason, the API response will be "UNKNOWN".

```

bigludo7: @jlurien I have an issue here because in location-retrieval we do not have unknown/UNKNOWN. 
We have either 

```
{
  "status": 400,
  "code": "LOCATION_RETRIEVAL.MAXAGE_INVALID_ARGUMENT",
  "message": "maxAge threshold cannot be satisfied"
}
```

or 

```
{
  "status": 404,
  "code": "NOT_FOUND",
  "message": "The specified resource is not found"
}
```

bigludo7 (commented):
Added @alpaycetin74 proposal for location-verification

bigludo7 (commented):
Add @alpaycetin74 proposal

code/API_definitions/location-retrieval.yaml
alpaycetin74: 400 seems more reasonable to me, as it points explicitly to the maxAge value.

bigludo7 (commented):
Remove space line 36

code/API_definitions/location-retrieval.yaml
bigludo7: Thanks @alpaycetin74 
Works for me and I can update the documentation accordingly.
If not able to provide require freshness (maxAge filled) --> `400` + code `LOCATION_RETRIEVAL.MAXAGE_INVALID_ARGUMENT`
If not able to provide any location for a request without maxAge --> `404`

jlurien: Right, I didn't realize that. 
Regarding errors, to me 400 is for cases when the provided input does not comply with the schema or some constraint that should be known by the client, independently of the device. It could fit for cases where some implementation does not support maxAge=0 for any user (it is still pending how to provide information about limits of the implementation to the client). But for cases where the maxAge cannot be satisfied for a specific device or at a specific moment, I think it is more of a  `404` or even  `422` (the entity combining device and maxAge is not processable by the system).

These RESTful interpretations of HTTP statuses are very subtle anyway.

bigludo7: OK - new proposal :)

(1)If  maxAge input does not fit the pattern (like `5fr/"3`)  --> 400
(2)If not able to locate the mobile --> 422 with code LOCATION_RETRIEVAL.UNABLE_TO_PROVIDE_LOCATION
(3)If not able to provide expected freshness -->  422 with a code LOCATION_RETRIEVAL.UNABLE_TO_PROVIDE_MAX_AGE

@jlurien @alpaycetin74

jlurien: For (2) I think 404 is more natural, the (3) is more about the combination of device + maxAge at a given time

(1)If maxAge input does not fit the pattern (like 5fr/"3) --> 400
(2)If not able to locate the mobile --> 404 with code LOCATION_RETRIEVAL.DEVICE_NOT_FOUND
(3)If not able to provide expected freshness --> 422 with a code LOCATION_RETRIEVAL.UNABLE_TO_FULFILL_MAX_AGE

cc @javier-carrocalabor

jlurien: Another difference with location-verification is that in this case we cannot return a lastLocationTime indicating how old is the more recent location

bigludo7 (commented):
Proposal following our discussion @jlurien @alpaycetin74

bigludo7 (commented):
Fixed megalinter issue

code/API_definitions/location-retrieval.yaml
alpaycetin74: Thanks @bigludo7 ,
if I may summarize my view:

For location retrieval,
(1) Obviously a client error, 400 is appropriate. 
It is possible to advise LOCATION_RETRIEVAL.MAXAGE_INVALID_ARGUMENT, but the implementer may not have full control over error messages created in case of syntactic errors - those may be created by software libraries. 
(2) No location available, even if maxAge is absent (infinite). 
404 with code LOCATION_RETRIEVAL.DEVICE_NOT_FOUND is appropriate.
(3) maxAge condition (including 0) cannot be satisfied. 
422 with code LOCATION_RETRIEVAL.UNABLE_TO_FULFILL_MAX_AGE is appropriate.

For location verification,
(1) same
(2) UNKNOWN without lastLocationTime
(3) UNKNOWN, may have  lastLocationTime if system has it.  
Thank you

bigludo7:
@alpaycetin74 @jlurien Kind reminder on this one. Is there still blocking point for you?

code/API_definitions/location-retrieval.yaml
```diff
@@ -37,6 +37,9 @@ info:
 
     * **Max Age**: Maximum age of the location information which is accepted for the location retrieval (in seconds).
 
+      * Absence of maxAge means "any age" is acceptable for the client. In other words, this is like maxAge=infinite. In this case the system may still return lastLocationTime, if available. If the system is not able to provide location a error 400 with code LOCATION_RETRIEVAL.DEVICE_NOT_FOUND is send back.
```
alpaycetin74, line 40: The text should refer to 404 instead of 400, I think.
I find 400 appropriate only if the maxAge is syntactically wrong.

code/API_definitions/location-retrieval.yaml
```diff
@@ -112,7 +115,9 @@ paths:
         '403':
           $ref: '#/components/responses/Generic403'
         '404':
-          $ref: '#/components/responses/Generic404'
+          $ref: '#/components/responses/RetrieveLocationNotFound404'
+        '422':
+          $ref: '#/components/responses/RetrieveLocationUnprocessableEntity404'
```
alpaycetin74, line 120: The schema needs to be renamed to RetrieveLocationUnprocessableEntity422

code/API_definitions/location-retrieval.yaml
```diff
@@ -393,8 +397,28 @@ components:
...
+              value:
+                status: 404
+                code: NOT_FOUND
+                message: 'The specified resource is not found'
+            deviceNotFound:
+              value:
+                status: 404
+                code: LOCATION_RETRIEVAL.DEVICE_NOT_FOUND
+                message: 'The location server is not able to locate the mobile'
+    RetrieveLocationUnprocessableEntity404:
```
alpaycetin74, line 420: The schema needs to be renamed to RetrieveLocationUnprocessableEntity422

bigludo7 (commented):
Fixed issues identified by @alpaycetin74

bigludo7 (commented):
Removal formula for UNABLE_TO_FULFILL_MAX_AGE

jlurien (approved):

PR #191, opened 2024-05-07
update base-path & security-scope for geofencing-subscriptions

maxl2287:
#### What type of PR is this?

Add one of the following kinds:
* correction

#### What this PR does / why we need it:
Allignment on https://github.com/camaraproject/Commonalities/pull/177. 

- [x] Rename to `geofencing-subscriptions.yaml` 
- [x] Rename basepath to `/geofencing-subscriptions/v0` 
- [x] update securityscope 

#### Which issue(s) this PR fixes:
Fixes #190 
Fixes #193

PR #195, opened 2024-05-17
Polygon area type added

### LINKED ISSUE #194: [Geofencing] Support for polygon area type
**Problem description**
Additionally to circles the Geofencing API should support polygon type areas. Circle type is rarely used as geofence. Usually a geofence describes a shape of an (often human built) geographical object like factories, buildings, city and country areas.

**Possible evolution**
Add polygon areas. The new datatype can be based on Geojson scheme (sorted array of 3DPoint-s) but using 2D geo type Point. 
Minimal number of points in the array is naturally set to 3. Maximal number can be restricted to 16 in order to avoid possible performance issues when calculating intersections between location area and geofence for triggering entering/leaving events.

```
    AreaType:
      type: string
      description: |
        Type of this area.
        POLYGON - The area is defined as polygon.
      enum:
        - CIRCLE
        - POLYGON

...

    Polygon:
      description: Polygon based area defined by an ordered geo points array
      type: object
      allOf:
        - $ref: "#/components/schemas/Area"
        - type: object
          properties:
            coordinates:
              type: array
              items:
                type: array
                items:
                minItems: 3
                maxItems: 16
      example:
        areaType: POLYGON
        coordinates:
          - latitude: 52.516770
            longitude: 13.378156
          - latitude: 52.516841
            longitude: 13.379433
          - latitude: 52.515917
            longitude: 13.37959
          - latitude: 52.515849
            longitude: 13.378308
```

ilya-smirnov-berlin:
#### What type of PR is this?

* enhancement/feature

Geofences should support polygon type areas. Circle type is rarely used as geofence. Usually a geofence describes a shape of an (often human built) geographical object like factories, buildings, city and country areas.

Fixes #194

#### Changelog input

```
Support for polygon area type

```


code/API_definitions/geofencing.yaml
```diff
@@ -342,6 +349,33 @@ components:
           longitude: 7.10066
         radius: 50000
 
+    Polygon:
```
jlurien, line 352: For coherence, we should use same schema that we already have defined for location-retrieval: https://github.com/camaraproject/DeviceLocation/blob/main/code/API_definitions/location-retrieval.yaml#L289, which is very similar, but uses `boundary` and `maxItems` is 15

ilya-smirnov-berlin: Thanks. The change is done.

code/API_definitions/geofencing.yaml
```diff
@@ -901,3 +938,48 @@ components:
...
+            boundary:
+              - latitude: 52.516770
+                longitude: 13.378156
+              - latitude: 52.516841
+                longitude: 13.379433
+              - latitude: 52.515917
+                longitude: 13.37959
+              - latitude: 52.515849
+                longitude: 13.378308
+          type: org.camaraproject.geofencing.v0.area-entered
```
alpaycetin74, line 961: Please bear in mind we have another PR https://github.com/camaraproject/DeviceLocation/pull/191 to rename security scopes and even types, so you may need to adapt. Thank you.

code/API_definitions/geofencing.yaml
jlurien: Thanks, we have to conclude the discussion in the issue #194  before deciding on the PR

code/API_definitions/geofencing.yaml
maxl2287: I recommend not updating the security scope and event types here until the #191 PR is merged. This can be done afterwards through merging and resolving any merge conflicts.

Otherwise, we will have "features" or "updates" that are not part of this PR/issue resolution. Additionally, if I update #191, we would also need to update it here.

jlurien:
Converted to Draft until we resolve the discussion in #194.

code/API_definitions/geofencing.yaml
jlurien: This feature is not in scope for next meta-release, so for now we keep this in Draft mode.

ilya-smirnov-berlin:
After rebasing to main lot of older commits are appear in PR. I can create a new PR or you can reopen and review this one.

maxl2287:
@ilya-smirnov-berlin please save your changes locally and create a new PR based on the current main.
It would help us a lot :)

PR #224, opened 2024-07-03
Add handling when the requested geofence `area` cannot be covered

### LINKED ISSUE #198: Geofencing - Manage when telco is not able to manage geofencing on the requested area
Another point that we should also probably discuss here is at subscription time.

Suppose we get a subscription for an area not covered or below a given percentage, or, for an area too small to be valid (N% will never be above x with current network ). 
In these case the consumer will never be able to receive notification.

I tend to think that we must be pro-active on this and refuse the subscription with a explicit message.  So we must add this capability in our API.

Probably this extra layer of checking could challenge our capability to answer synchronously to the call so also I'm wondering if we should not add capability to answer asynchronously to geofencing request.

_Originally posted by @bigludo7 in https://github.com/camaraproject/DeviceLocation/issues/133#issuecomment-2072579733_

maxl2287:
#### What type of PR is this?

* enhancement/feature

Adds handling when the provided subscription area cannot be covered or is too small to receive any notifications.

Fixes #198

### DIFF
--- a/code/API_definitions/geofencing-subscriptions.yaml
+++ b/code/API_definitions/geofencing-subscriptions.yaml
@@ -913,7 +913,7 @@ components:
       type: string
       description: |
         - NETWORK_TERMINATED - API server stopped sending notification
-        - SUBSCRIPTION_CREATION_ABORTED - Subscription cannot be created because the specified area cannot be managed (for asynchronous subscription creation) 
+        - SUBSCRIPTION_CREATION_ABORTED - Subscription cannot be created because the specified area cannot be managed (for asynchronous subscription creation)
         - SUBSCRIPTION_EXPIRED - Subscription expire time (optionally set by the requester) has been reached
         - MAX_EVENTS_REACHED - Maximum number of events (optionally set by the requester) has been reached
         - ACCESS_TOKEN_EXPIRED - Access Token sinkCredential (optionally set by the requester) expiration time has been reached
--- a/code/API_definitions/geofencing-subscriptions.yaml
+++ b/code/API_definitions/geofencing-subscriptions.yaml
@@ -1106,11 +1106,6 @@ components:
                 status: 400
                 code: "INVALID_TOKEN"
                 message: "Only bearer token is supported"
-            AreaNotCovered:
-              value:
-                status: 400
-                code: "AREA_NOT_COVERED"
-                message: "The specified area cannot be covered or is too small to be valid"
 
     Generic400:
       description: Problem with the client request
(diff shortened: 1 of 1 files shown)


bigludo7 (dismissed):
LGTM

maxl2287:
@bigludo7 do you align with the TerminationReason?

bigludo7:
> @bigludo7 do you align with the TerminationReason?

I've raised a generic issue in commonalities ([243](https://github.com/camaraproject/Commonalities/issues/243)). Here we have a specific one. I'm not sure about commonalities outcome so probably we can keep this one open to get team feedback, and in // check if we have a global direction from commonalities.

maxl2287:
In DeviceStatus we have a label `waiting for commonalities`.
Maybe we can also use such label here 🤔

code/API_definitions/geofencing-subscriptions.yaml
```diff
@@ -912,12 +913,14 @@ components:
       type: string
       description: |
         - NETWORK_TERMINATED - API server stopped sending notification
+        - SUBSCRIPTION_CREATION_ABORTED - Subscription cannot be created because the specified area cannot be managed (for asynchronous subscription creation)
```
jlurien, line 916: If the reason is generic I would list the area case as an example, it may be others. Otherwise we should create a dedicated termination reason for it. If the client needs to know that the area is invalid we may need to inform of that specifically.

```suggestion
        - SUBSCRIPTION_CREATION_ABORTED - Subscription cannot be created (relevant for asynchronous subscription creation). For example, because the specified area cannot be managed 
```

maxl2287: In Commonalities we could use the following property:

```yaml
terminationDescription: 
   type: string 
```

maxl2287: https://github.com/camaraproject/Commonalities/blob/main/artifacts/camara-cloudevents/event-subscription-template.yaml#L629-L630

code/API_definitions/geofencing-subscriptions.yaml
```diff
@@ -923,15 +925,15 @@ components:
       type: string
       description: |
         - NETWORK_TERMINATED - API server stopped sending notification
-        - SUBSCRIPTION_CREATION_ABORTED - Subscription cannot be created because the specified area cannot be managed (for asynchronous subscription creation)
+        - SUBSCRIPTION_UNPROCESSABLE - Subscription cannot be processed because the specified area cannot be managed (for asynchronous subscription creation)
```
jlurien, line 928:
```suggestion
        - SUBSCRIPTION_UNPROCESSABLE - Subscription cannot be processed due to some reason, e.g. because the specified area cannot be managed. Useful for asynchronous subscription creation.
```

It can be generalized to any reason. Termination description can provide the details.

jlurien (dismissed):
Approved with a suggestion

bigludo7 (dismissed):
LGTM with José suggestion

jlurien (approved):
LGTM

PR #119, opened 2023-11-20
Add Test Definition for location Retrieval

bigludo7:
In order to prepare a release candidate added Test Definition for location Retrieval

#### What type of PR is this?

* tests

#### What this PR does / why we need it:

In order to prepare a release candidate added Test Definition for location Retrieval

#### Which issue(s) this PR fixes:

Fixes #

#### Special notes for reviewers:

To be done consistently with https://github.com/camaraproject/Commonalities/issues/94

#### Changelog input

```
 release-note

```

#### Additional documentation 

This section can be blank.

```
docs

```

PR #229, opened 2024-07-16
Test definition location retrieval

### LINKED ISSUE #196: Provide Test Defintion for Device Retrieval API
**Problem description**
Provide Test Defintion for Device Retrieval API

**Expected action**
Provide Gherkin test definition file

**Additional context**

bigludo7:
#### What type of PR is this?

* tests

Provide test definition for Location-retrieval

Fixes #196 

I got an issue with my previous PR #119 so I have to create a new one :(

#### Changelog input

```
 release-note

```

#### Additional documentation 

This section can be blank.

```
docs

```


code/Test_definitions/Location-Retrieval.feature
```diff
@@ -0,0 +1,264 @@
...
+  Scenario: None of the provided device identifiers is supported by the implementation
+    Given that some type of device identifiers are not supported by the implementation
+    And the request body property "$.device" only includes device identifiers not supported by the implementation
+    When the HTTP "POST" request is sent
+    Then the response status code is 422
+    And the response property "$.status" is 422
+    And the response property "$.code" is "UNSUPPORTED_DEVICE_IDENTIFIERS"
+    And the response property "$.message" contains a user friendly text
+
+  @location_retrieval_13_device_not_found
```
trehman-gsma, line 148: Will this 404 case be possible in three-legged token use cases? The access token would be tied to a valid device already?

code/Test_definitions/Location-Retrieval.feature
```diff
@@ -0,0 +1,264 @@
...
+    Given the request body is set to "{}"
+    When the HTTP "POST" request is sent
+    Then the response status code is 400
+    And the response property "$.status" is 400
+    And the response property "$.code" is "INVALID_ARGUMENT"
+    And the response property "$.message" contains a user friendly text
+
+  # Other specific 400 errors
+
+  @location_retrieval_400.3_max_age_schema_compliant
```
trehman-gsma, line 211: What do you think about one more test case to validate the maxAge minimum value compliance (minimum=60)?

It could either be combined with this test case (@location_retrieval_400.3_max_age_schema_compliant) or it could be a new test case. Examples below.

Combining into one:
```
@location_retrieval_400.3_max_age_schema_compliant
Scenario: maxAge value does not comply with the schema
    Given the request body property "$.device" is set to a valid testing device
    And the request body property "maxAge" does not comply with the OAS schema at "/components/schemas/RetrievalLocationRequest" 
    When the HTTP "POST" request is sent
    Then the response status code is 400
    And the response property "$.status" is 400
    And the response property "$.code" is "INVALID_ARGUMENT"
    And the response property "$.message" contains a user friendly text

    Examples:
      | maxAge  |
      | 6a0     |
      | 59      |
```

Creating a separate use case:

```
@location_retrieval_400.4_max_age_min_schema_compliant
Scenario: maxAge value is below the minimum (60) allowed
    Given the request body property "$.device" is set to a valid testing device
    And the "maxAge" is set to 59
    When the HTTP "POST" request is sent
    Then the response status code is 400
    And the response property "$.status" is 400
    And the response property "$.code" is "INVALID_ARGUMENT"
    And the response property "$.message" contains a user friendly text

```

code/Test_definitions/Location-Retrieval.feature
```diff
@@ -0,0 +1,264 @@
...
+    Given the request body property "$.device" is set to a valid testing device
+    And the "maxAge" is set to 6a0
+    When the HTTP "POST" request is sent
+    Then the response status code is 400
+    And the response property "$.status" is 400
+    And the response property "$.code" is "INVALID_ARGUMENT"
+    And the response property "$.message" contains a user friendly text
+
+
+  @location_retrieval_400.4_required_device_identifier_missing
```
trehman-gsma, line 222: Will device remain a mandatory field in the next release?
(Referring to three legged use case where access token is tied to a device)

code/Test_definitions/Location-Retrieval.feature
bigludo7: @trehman-gsma we have removed the maxAge limitation. You can omit it or pass it with 0 value as well.

code/Test_definitions/Location-Retrieval.feature
bigludo7: This is a fair question.
By the generic rule this API will require 3-legs but what if this API is used to track device (not linked to people) for a B2B customer? I tend to think we can have some borderline UC in 2-legs. I prefer to keep it as optional but if we prefer to remove it I understand.
@jlurien WDYT?

code/Test_definitions/Location-Retrieval.feature
trehman-gsma: oh! I must have been looking at an older spec then, apologies!

code/Test_definitions/Location-Retrieval.feature
jlurien: If we make device optional this test case would need to be adapted. For 2-legged access,  or even for 3-legged when a single device cannot be deducted from the access token, it would be an error, but not 400 INVALID_ARGUMENT (not compliant with schema)  but 422 UNIDENTIFIABLE_DEVICE

code/Test_definitions/Location-Retrieval.feature
jlurien: There may be situations as @bigludo7 suggests when the access token is valid but the device included in the request does not point to any device. 

* Device can NOT be deducted from the access token:
  * `device` NOT included in the request body-> 422 UNIDENTIFIABLE_DEVICE
  * `device` included in the request body but not pointing to a known device -> 404 DEVICE_NOT_FOUND

jlurien (approved):
OK with the clarifications, We may need to adapt some scenarios if we make device opcional and add support for 422 UNIDENTIFIABLE_DEVICE, but we make that changes along with the adaptations in the API specs

PR #242, opened 2024-07-29
geofencing: Add `SUBSCRIPTION_DELETED` as termination-reason

### LINKED ISSUE #141: Geofencing - Adding `SUBSCRIPTION_DELETED` as a Termination Reason

maxl2287:
#### What type of PR is this?

* enhancement/feature

Adds a new termination-reason for "subscription-ends" - CloudEvent, when the user deletes the subscriptions via 
`DELETE /subscriptions/{subscriptionId}`

Fixes #141

### DIFF
--- a/code/API_definitions/geofencing-subscriptions.yaml
+++ b/code/API_definitions/geofencing-subscriptions.yaml
@@ -47,6 +47,7 @@ info:
     It is used when:
       - the subscription expire time (optionally set by the requester) has been reached
       - the maximum number of subscription events (optionally set by the requester) has been reached
+      - the subscription was deleted by the requester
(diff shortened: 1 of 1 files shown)


jlurien (approved):
LGTM

PR #245, opened 2024-07-30
Latest alignments and fixes for verification and retrieval

jlurien:
#### What type of PR is this?

* correction
* cleanup
* documentation

Aligns both APIs between them
Adds note about networkAccessIdentifier not to be used with this version of guidelines
contact info not to be included
license info to be included

Fixes #

#### Changelog input

```
 release-note

```

#### Additional documentation 

This section can be blank.

```
docs

```

### DIFF
--- a/code/API_definitions/location-retrieval.yaml
+++ b/code/API_definitions/location-retrieval.yaml
@@ -13,7 +13,7 @@ info:
 
     The retrieved shape depends on the network conditions at the subscriber's location and any of the supported shapes could be received.
 
-    The requester could optionally ask for a freshness of the localization information by providing a maxAge ("I want a location not older than 600 seconds").
+    The requester could optionally ask for a freshness of the localization information by providing a `maxAge` ("I want a location not older than 600 seconds").
 
     The result accuracy depends on the network's ability and accuracy to locate the device.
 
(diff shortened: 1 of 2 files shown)


jlurien:
@maxl2287 please take a look and approve if OK. Please double check if you need to apply some the same fixes to geofencing

maxl2287 (approved):
LGTM.
Short question: Is this optional device also enforced for subscriptions?

maxl2287:
> LGTM. Short question: Is this optional device also enforced for subscriptions?

Okay on DeviceStatus it's also still inside the subscriptionDetail to be mandatory.

https://github.com/camaraproject/DeviceStatus/blob/f089994491c7aac658e434c05e09873a473e2d68/code/API_definitions/device-roaming-status-subscriptions.yaml#L506C1-L514C17

jlurien:
> LGTM. Short question: Is this optional device also enforced for subscriptions?

No, for geofencing we decided in a past meeting to keep it as required, because it is returned in the response. One consideration we have been discussing in QoD is to remark that the identifiers that are returned for the device in the response must be the same values that were passed as device identifiers in the request (in our case during subscription creation). This is to avoid disclosing unnecessary and potentially sensible info to the client (e.g. receive an IP as identifier and return the phone number of the device)

PR #246, opened 2024-07-31
Update externalDocs for rc.1

jlurien:
#### What type of PR is this?

* correction

Align specs with review guidelines


jlurien:
@maxl2287, @hdamker I need to merge this one before creating the final PR fo r1.1 which is already drafted (#247)

PR #262, opened 2024-10-01
Update location-retrieval with add `maxSurface` management

### LINKED ISSUE #255: How to understand the response parameter area in the device retrieval API
**Problem description**
Hi Teams,
About the location retrieval API, we are currently implementing this API and have encountered a problem. Regarding the region in the response parameters (including the type and size of the region), how is it obtained? Is it defined by the operator themselves? Are there any related constraints (such as not allowing the return area to be too large) and whether the actual location of device needs to be within the return area.

bigludo7:
#### What type of PR is this?

* enhancement/feature

add `maxSurface` management

Fixes #255 

#### Changelog input

```
 release-note
- Add management of maxSurface in request.

```

#### Additional documentation 

This section can be blank.

```
docs

```

### DIFF
--- a/code/API_definitions/location-retrieval.yaml
+++ b/code/API_definitions/location-retrieval.yaml
@@ -13,7 +13,9 @@ info:
 
     The retrieved shape depends on the network conditions at the device's location and any of the supported shapes could be received.
 
-    The requester could optionally ask for a freshness of the localization information by providing a `maxAge` ("I want a location not older than 600 seconds").
+    The requester could optionally ask for
+     * a freshness of the localization information by providing a `maxAge` ("I want a location not older than 600 seconds").
+     * an accuracy of the localization information by providing a `maxSurface` ("I want a location not larger than 1000000 meters square").
 
     The result accuracy depends on the network's ability and accuracy to locate the device.
 
@@ -41,6 +43,11 @@ info:
       * `maxAge`=0 means that a fresh calculation is requested by the client. If the system is not able to provide the fresh location, an error 422 with code LOCATION_RETRIEVAL.UNABLE_TO_FULFILL_MAX_AGE is sent back.
 
     * **Last Location Time** : Last date and time when the device was localized.
+
+    * **Max Surface**: Maximum surface in square meters which is accepted by the client for the location retrieval.
+
+      * absence of `maxSurface` means that "any surface size" is acceptable for the client. 
+      * If the system is not able to provide an area with a surface acceptable with the client request, an error 422 with code LOCATION_RETRIEVAL.UNABLE_TO_FULFILL_MAX_SURFACE is sent back.
 
 
     # API Functionality
@@ -87,7 +94,7 @@ info:
     # Further info and support
 
     (FAQs will be added in a later version of the documentation)
-  version: 0.3.0
+  version: wip
   license:
     name: Apache 2.0
     url: https://www.apache.org/licenses/LICENSE-2.0.html
@@ -96,7 +103,7 @@ externalDocs:
   description: Project documentation at Camara
   url: https://github.com/camaraproject/DeviceLocation
 servers:
-  - url: '{apiRoot}/location-retrieval/v0.3'
+  - url: '{apiRoot}/location-retrieval/wip'
     variables:
       apiRoot:
         default: http://localhost:9091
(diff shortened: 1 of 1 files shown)


code/API_definitions/location-retrieval.yaml
```diff
@@ -525,6 +545,13 @@ components:
                 status: 422
                 code: LOCATION_RETRIEVAL.UNABLE_TO_FULFILL_MAX_AGE
                 message: "Unable to provide expected freshness for location"
+            LOCATION_RETRIEVAL_422_UNABLE_TO_FULFILL_MAX_SURFACE:
```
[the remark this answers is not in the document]
bigludo7: Good point. I've added a note: if both `maxAge` and `maxSurface` requirements fail, the system can either send back one or the other error code. 

I agree on your recommendation but not sure if we have to add in the yaml as it is more for the system than the client.

code/API_definitions/location-retrieval.yaml
```diff
@@ -13,7 +13,9 @@ info:
 
     The retrieved shape depends on the network conditions at the device's location and any of the supported shapes could be received.
 
-    The requester could optionally ask for a freshness of the localization information by providing a `maxAge` ("I want a location not older than 600 seconds").
+    The requester could optionally ask for
+     * a freshness of the localization information by providing a `maxAge` ("I want a location not older than 600 seconds").
+     * an accuracy of the localization information by providing a `maxSurface` ("I want a location not larger than 1000000 meters square").
```
jlurien, line 18:
```suggestion
     * an accuracy of the localization information by providing a `maxSurface` ("I want a location not larger than 1000000 square meters").

```

code/API_definitions/location-retrieval.yaml
```diff
@@ -203,7 +220,11 @@ components:
         maxAge:
           type: integer
           description: Maximum age of the location information which is accepted for the location retrieval (in seconds). Absence of maxAge means "any age" and maxAge=0 means a fresh calculation.
-
+        maxSurface:
+          type: integer
+          minimum: 1
```
jlurien, line 225: Not sure what would be realistically the minimum that a "best" implemention may provide, but likely above 100 m2. We may set 1 here and let implementations provide the minimum supported separately.

jlurien (approved):
LGTM with minor correction and comment

code/API_definitions/location-retrieval.yaml
bigludo7: As we allow a min radius of 1 probably we cannot have a min of 100m2. Probably good to add an note indicating that implementations will provide minimum.

code/API_definitions/location-retrieval.yaml
```diff
@@ -96,7 +106,7 @@ externalDocs:
   description: Project documentation at Camara
   url: https://github.com/camaraproject/DeviceLocation
 servers:
-  - url: '{apiRoot}/location-retrieval/v0.3'
+  - url: '{apiRoot}/location-retrieval/wip'
```
jlurien, line 109:
```suggestion
  - url: '{apiRoot}/location-retrieval/vwip'
```

code/API_definitions/location-retrieval.yaml
```diff
@@ -42,6 +45,13 @@ info:
 
     * **Last Location Time** : Last date and time when the device was localized.
 
+    * **Max Surface**: Maximum surface in square meters which is accepted by the client for the location retrieval.
+
+      * absence of `maxSurface` means that "any surface size" is acceptable for the client.
+      * API implementation could specify in the documentation the `maxSurface` minimum acceptable (for exemple minimun of 10000 square meters allowed).
```
alpaycetin74, line 51: I'd like to propose a minor rewording here, and fixing some typos:
API implementation could specify the minimum acceptable `maxSurface`  in the documentation (for example a minimum of 10000 square meters are allowed).

jlurien, line 51:
```suggestion
      * API implementation could specify the minimum acceptable `maxSurface` in the documentation (for example a minimum of 10000 square meters are allowed).
```

jlurien (dismissed):
With suggestion

jlurien (approved):

Issue #277, opened 2024-12-08
Update info.description with commonalities


bigludo7:
**Problem description**
For Location retrieval & Location Verification the info.description must bu updated accordingly to https://github.com/camaraproject/Commonalities/blob/main/documentation/API-design-guidelines.md#appendix-a-normative-infodescription-template-for-when-user-identification-can-be-from-either-an-access-token-or-explicit-identifier 

Location Retrieval - lines 63 to 85
Location Verification - lines 56-78

**Expected behavior**

**Alternative solution**

**Additional context**
Up to my understanding this update is not required for Geofencing.

maxl2287:
These changes were done by #281 (location-verification) and #283 (location-retrieval).

cc @bigludo7 & @jlurien

PR #281, opened 2024-12-17
Update location-verification.yaml for Spring25

### LINKED ISSUE #276: Error Alignement with Commonalities
**Problem description**
Following Commonalities rules:

To be applied:
Error 429 must be added for location-retrieval (already present in location-verification & geofencing)
Errors 410 & 429 must be added to the callback endpoint in Geofencing
Error 409 must be added for POST /subscriptions in Geofencing

To be discussed:
Error code 500 & 503 could be removed but we need a decision for all 3 APIs for consistency
Error code 429 could be removed for GET & DELETE subscriptions in Geofencing

**Expected behavior**
Update yamls

**Alternative solution**

**Additional context**

jlurien:
#### What type of PR is this?

* enhancement/feature

Updates location-verification API with the changes derived from Commonalities and ICM updates
x-camara-guidelines: 0.5.0

Fixes #276, #277 

There are still some open discussions in Commonalities that may impact this PR, so we will keep it open until there is a final version for x-camara-guidelines: 0.5.0

#### Changelog input

```
* Error schemas updated with enums
* Some error status and codes have been updated
* Section with guidelines about how to identify the device from access token and request body has been updated 
```

### DIFF
--- a/code/API_definitions/location-verification.yaml
+++ b/code/API_definitions/location-verification.yaml
@@ -494,7 +494,6 @@ components:
                   code:
                     enum:
                       - UNAUTHENTICATED
-                      - AUTHENTICATION_REQUIRED
           examples:
             GENERIC_401_UNAUTHENTICATED:
               description: Request cannot be authenticated
@@ -502,12 +501,6 @@ components:
                 status: 401
                 code: UNAUTHENTICATED
                 message: Request not authenticated due to missing, invalid, or expired credentials.
-            GENERIC_401_AUTHENTICATION_REQUIRED:
-              description: New authentication is needed, authentication is no longer valid
-              value:
-                status: 401
-                code: AUTHENTICATION_REQUIRED
-                message: New authentication is required.
 
     Generic403:
       description: Forbidden
@@ -527,7 +520,6 @@ components:
                   code:
                     enum:
                       - PERMISSION_DENIED
-                      - INVALID_TOKEN_CONTEXT
           examples:
             GENERIC_403_PERMISSION_DENIED:
               description: Permission denied. OAuth2 token access does not have the required scope or when the user fails operational security
@@ -535,12 +527,6 @@ components:
                 status: 403
                 code: PERMISSION_DENIED
                 message: Client does not have sufficient permissions to perform this action.
-            GENERIC_403_INVALID_TOKEN_CONTEXT:
-              description: Reflect some inconsistency between information in some field of the API and the related OAuth2 Token
-              value:
-                status: 403
-                code: INVALID_TOKEN_CONTEXT
-                message: "{{field}} is not consistent with access token."
 
     Generic404:
       description: Not found
--- a/code/API_definitions/location-verification.yaml
+++ b/code/API_definitions/location-verification.yaml
@@ -325,7 +325,7 @@ components:
         * `networkAccessIdentifier`
 
         NOTE1: the MNO might support only a subset of these options. The API invoker can provide multiple identifiers to be compatible across different MNOs. In this case the identifiers MUST belong to the same device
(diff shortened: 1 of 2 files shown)


code/API_definitions/location-verification.yaml
```diff
@@ -53,29 +53,19 @@ info:
...
 
-    # Identifying a device from the access token
+    # Identifying the device from the access token
 
-    This specification defines the `device` object field as optional in API requests, specifically in cases where the API is accessed using a 3-legged access token, and the device can be uniquely identified by the token. This approach simplifies API usage for API consumers by relying on the device information associated with the access token used to invoke the API.
+    This API requires the API consumer to identify a device as the subject of the API as follows:
+    - When the API is invoked using a two-legged access token, the subject will be identified from the optional `device` object, which therefore MUST be provided.
+    - When a three-legged access token is used however, this optional identifier MUST NOT be provided, as the subject will be uniquely identified from the access token.
```
alpaycetin74, line 60: If both the 3-legged token and the device is provided, will the request be simply rejected? Is it possible to let the request continue if the token and device match?

code/API_definitions/location-verification.yaml
```diff
@@ -489,17 +484,25 @@ components:
...
             GENERIC_401_UNAUTHENTICATED:
-              summary: Generic Unauthenticated
               description: Request cannot be authenticated
               value:
                 status: 401
                 code: UNAUTHENTICATED
                 message: Request not authenticated due to missing, invalid, or expired credentials.
             GENERIC_401_AUTHENTICATION_REQUIRED:
-              summary: Generic Authentication Required
               description: New authentication is needed, authentication is no longer valid
```
alpaycetin74, line 506: is it possible to explain the difference between "authentication is no longer valid" and "expired credentials" in more detail ?

bigludo7 (dismissed):
LGTM

code/API_definitions/location-verification.yaml
jlurien: There is an issue in Commonalities to discuss this: https://github.com/camaraproject/Commonalities/issues/368

IMO, UNAUTHENTICATED is the most generic one, which would cover all cases where the access token is not longer valid (typically because it is expired). AUTHENTICATION_REQUIRED is more specific, when a full new authentication from scratch is needed. But we could use the generic one and rely on the message for further description.

jlurien (commented):
- Test plan should be updated in the same PR

jlurien:
@bigludo7 @maxl2287 

I have updated the PR with new changes:

* Updated the test plan. For that I have copied the **fixed** version of the test plan artifact in Commonalities, now still proposed as [PR](https://github.com/camaraproject/Commonalities/pull/381), but likely to be merged. We would probably need to apply as well to location-retrieval and check what applies to geofencing

* From the spec, I have already removed 403 INVALID_TOKEN_CONTEXT, and 401 AUTHENTICATION_REQUIRED. Discussions around this are not fully closed but it seems this will be the consensus.

code/Test_definitions/location-verification.feature
```diff
@@ -115,70 +117,82 @@ Feature: CAMARA Device location verification API, v1.0.0 - Operation verifyLocat
...
+    And the response property "$.status" is 422
+    And the response property "$.code" is "MISSING_IDENTIFIER"
+    And the response property "$.message" contains a user-friendly text
+
+
+  @location_verification_C01.06_unsupported_device
+  Scenario: None of the provided device identifiers is supported by the implementation
+    Given that some types of device identifiers are not supported by the implementation
+    And the request body property "$.device" only includes device identifiers not supported by the implementation
+    When the HTTP "POST" request is sent
```
bigludo7, line 163: We should probably add: And the header "Authorization" is set to a valid access token which does not identify a single device

bigludo7 (commented):
One small comment on the test definition else good for me.

code/Test_definitions/location-verification.feature
jlurien: Yes, you are right, but this is copied from Commonalities artifact, so I will have to update there as well

maxl2287 (approved):

bigludo7 (approved):
LGTM

PR #291, opened 2025-01-23
Add MultiSIM section to info.description

### LINKED ISSUE #257: Support of Multi-SIM lines in DeviceLocation APIs
**Problem description**
As introduced in https://github.com/camaraproject/Commonalities/issues/302, for Multi-SIM services. providing a phoneNumber as input does not identify a single device uniquely.

Current DeviceLocation specs do not consider or give guidelines about how to behave in scenarios when the input is only a phoneNumber of a Multi-SIM service and specific device (SIM) cannot be inferred from the access token or other device identifier.
 
**Possible evolution**
Discuss the problem and alternatives. There may be impact in other tracks, such as IC&M. We may need also to discuss implications in the short term when dealing with this situation in current versions.

In the case of DeviceLocation each API has each own considerations, as each SIM may be located in a different place:

* For location-verification, the input area can match the location of one of the SIMs but not others. Do we return TRUE if there is a single match, or PARTIAL or another new value?
* For location-retrieval, do we have to return an array with several locations? Do we want to disclose that the client has multi-SIM? Would we need to identify each device?
* For geofencing-subscriptions? Do we trigger an event when any of the SIMs enter or leaves the area? How do we identify which device/SIM is the one affected?

**Alternative solution**
- Relying on 3-legged tokens as much as possible would minimize the problem, but currently we cannot assure that they always identify a single device instead of a phone number. 
- Finally we can handle the situation with errors, such as 422 UNIDENTIFIABLE_DEVICE, or another dedicated code, informing clients to provide an alternative identifier that can identify the specific device. But this should be limited to corner situations as it is not dev friendly and in most cases developers do not know if a phone number belongs to a multi-SIM service.

jlurien:
#### What type of PR is this?

* documentation

Current documentation omits the potential problem with phone numbers identifying a MultiSIM group. As there is no a single behaviour agreed in CAMARA yet, the proposed enhancement presents the problem and gives some hints to mitigate it

Fixes #257

#### Changelog input

```
* Adding section for "Multi-SIM scenario handling" in info.description
```

### DIFF
--- a/code/API_definitions/geofencing-subscriptions.yaml
+++ b/code/API_definitions/geofencing-subscriptions.yaml
@@ -81,7 +81,7 @@ info:
 
     - Using preferably the authorisation code flow to obtain an access token, which will automatically identify the intended device.
     - Identifying the intended device from a unique identifier for that device, such as its source IP address and port.
-    - Check with the SIM provider whether a unique "secondary" phone number is already associated with each device and use the secondary phone number to identify the intended device if available.
+    - Check with the API provider whether a unique "secondary" phone number is already associated with each device and use the secondary phone number to identify the intended device if available.
 
     # Further info and support
 
--- a/code/API_definitions/geofencing-subscriptions.yaml
+++ b/code/API_definitions/geofencing-subscriptions.yaml
@@ -75,14 +75,14 @@ info:
 
     - rejecting the subscription with an error indicating that that phone number is not supported for this API, or
     - generating events for a single device in the multi-SIM group, if one of the devices is considered linked to the main SIM and this concept is supported by the operator, or
-    - generating events for of all the SIMs associated to the requested phone number. 
-    
+    - generating events for of all the SIMs associated to the requested phone number.
+
     Possible solutions to make the scenario more deterministic include:
 
     - Using preferably the authorisation code flow to obtain an access token, which will automatically identify the intended device.
     - Identifying the intended device from a unique identifier for that device, such as its source IP address and port.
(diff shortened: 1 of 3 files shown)


code/API_definitions/geofencing-subscriptions.yaml
```diff
@@ -69,6 +69,20 @@ info:
...
+
+    - rejecting the subscription with an error indicating that that phone number is not supported for this API, or
+    - generating events for a single device in the multi-SIM group, if one of the devices is considered linked to the main SIM and this concept is supported by the operator, or
+    - generating events for of all the SIMs associated to the requested phone number.
+
+    Possible solutions to make the scenario more deterministic include:
+
+    - Using preferably the authorisation code flow to obtain an access token, which will automatically identify the intended device.
+    - Identifying the intended device from a unique identifier for that device, such as its source IP address and port.
+    - Check with the SIM provider whether a unique "secondary" phone number is already associated with each device and use the secondary phone number to identify the intended device if available.
```
alpaycetin74, line 84: In line 74 we use the term API provider, but here we use SIM provider. Could we use a common term like service provider or opco ? Thank you.

maxl2287 (dismissed):
LGTM

code/API_definitions/geofencing-subscriptions.yaml
jlurien: Adjusted

maxl2287 (approved):

bigludo7 (approved):
LGTM
Thanks Jose !

PR #298, opened 2025-02-15
Set Device Object To Optional - According to API Design Guidelines

### LINKED ISSUE #297: Device Object Should Be Optional According to API Design Guidelines
**Problem description**
In the current specification (geofencing-subscriptions.yaml), the Device object is defined as a mandatory property within SubscriptionDetail. This would consistently result in error 422 UNNECESSARY_IDENTIFIER with the message: "The device is already identified by the access token." whenever the subject is also identified by the access token.
```yaml
    SubscriptionDetail:
      description: The detail of the requested event subscription.
      type: object
      properties:
        device:
          $ref: "#/components/schemas/Device"
        area:
          $ref: "#/components/schemas/Area"
      required:
        - device
        - area
```

**Expected behavior**
To prevent this issue and ensure compliance with the guidelines, the Device object should be made optional.
```yaml
    SubscriptionDetail:
      description: The detail of the requested event subscription.
      type: object
      properties:
        device:
          $ref: "#/components/schemas/Device"
        area:
          $ref: "#/components/schemas/Area"
      required:
        - area
```

dfischer-tech:
#### What type of PR is this?

* correction

Fixes #297 

#### Changelog input

```
 release-note
 Update geofencing-subscriptions.yaml to make the Device object optional.

```

#### Additional documentation 

None

### DIFF
--- a/code/API_definitions/geofencing-subscriptions.yaml
+++ b/code/API_definitions/geofencing-subscriptions.yaml
@@ -72,6 +72,12 @@ info:
 
     In cases where personal data is processed by the API and users can exercise their rights through mechanisms such as opt-in and/or opt-out, the use of three-legged access tokens is mandatory. This ensures that the API remains in compliance with privacy regulations, upholding the principles of transparency and user-centric privacy-by-design.
 
+    # Identifying the device from the access token
+
+    This API requires the API consumer to identify a device as the subject of the API as follows:
+    - When the API is invoked using a two-legged access token, the subject will be identified from the optional `device` object, which therefore MUST be provided.
+    - When a three-legged access token is used however, this optional identifier MUST NOT be provided, as the subject will be uniquely identified from the access token.
+
     # Multi-SIM scenario handling
 
     In multi-SIM scenarios, where more than one mobile device is associated with the phone number given as input in the API call (e.g. a smartphone with an associated smartwatch), it might not be possible to uniquely identify the device whose location is to be verified. Check with the API provider what is the expected behaviour when a phone number belonging to a multi-SIM group is used as the device identifier, as the API may behave:
--- a/code/API_definitions/geofencing-subscriptions.yaml
+++ b/code/API_definitions/geofencing-subscriptions.yaml
@@ -879,10 +879,13 @@ components:
               $ref: "#/components/schemas/SubscriptionEnds"
 
     AreaLeft:
-      description: Event detail structure for area-left event.
+      description: |
+        Event detail structure for org.camaraproject.geofencing-subscriptions.v0.area-left event.
+        Note that the device object is defined as optional and will only to be returned if provided in createSubscription. 
+        If more than one type of device identifier was provided, only one identifier will be returned 
+        (at implementation choice and with the original value provided in createSubscription).
       type: object
       required:
-        - device
         - area
         - subscriptionId
       properties:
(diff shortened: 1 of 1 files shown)


bigludo7 (dismissed):
LGTM

maxl2287:
Thanks @dfischer-tech for raising up this PR!
Could you please also update the `info.description` by adding the information about how to identify the device from the access-token:

```yaml
    # Identifying the device from the access token

    This API requires the API consumer to identify a device as the subject of the API as follows:
    - When the API is invoked using a two-legged access token, the subject will be identified from the optional `device` object, which therefore MUST be provided.
    - When a three-legged access token is used however, this optional identifier MUST NOT be provided, as the subject will be uniquely identified from the access token.
```
(see: https://github.com/camaraproject/DeviceLocation/blob/5e81e4e6dde3f249bba9a8ff69ddc922475dced4/code/API_definitions/location-retrieval.yaml#L73-L77)

maxl2287 (changes requested):
Thanks @dfischer-tech for raising up this PR!
Could you please also update the `info.description` by adding the information about how to identify the device from the access-token:

```yaml
    # Identifying the device from the access token

    This API requires the API consumer to identify a device as the subject of the API as follows:
    - When the API is invoked using a two-legged access token, the subject will be identified from the optional `device` object, which therefore MUST be provided.
    - When a three-legged access token is used however, this optional identifier MUST NOT be provided, as the subject will be uniquely identified from the access token.

    This approach simplifies API usage for API consumers using a three-legged access token to invoke the API by relying on the information that is associated with the access token and was identified during the authentication process.

```
(see: https://github.com/camaraproject/DeviceLocation/blob/5e81e4e6dde3f249bba9a8ff69ddc922475dced4/code/API_definitions/location-retrieval.yaml#L73-L77)

maxl2287 (dismissed):
LGTM

jlurien (commented):
If we opted to remove device as required in the request, we would need to remove it as well in the event contents. schemas `AreaLeft`, `AreaEnetered`, `SubscriptionEnds`, and probably add some guideline that `device` must not be included in the event if not explicitly provided in the request

dfischer-tech:
> If we opted to remove device as required in the request, we would need to remove it as well in the event contents. schemas `AreaLeft`, `AreaEntered`, `SubscriptionEnds`, and probably add some guideline that `device` must not be included in the event if not explicitly provided in the request

Hi @jlurien,
Thank you for your review and comment.
The device is set to "optional" to prevent conflicts when it has already been identified via an access token (which would otherwise result in a 422 - UNNECESSARY_IDENTIFIER).
However, the device still needs to be identified; otherwise, a 422 - MISSING_IDENTIFIER error would occur, with the message:

"An identifier is not included in the request and the device or phone number identification cannot be derived from the 3-legged access token."

Therefore, we do not need to explicitly remove it from the event contents (schemas: AreaLeft, AreaEntered, SubscriptionEnds),
because it must be there/identified.

maxl2287:
@dfischer-tech 

For a better understanding I would like to reference here as well to QualityOnDemand:

>Note that the device object is defined as optional and will only to be returned if provided in createSession.
(https://github.com/camaraproject/QualityOnDemand/blob/r2.1/code/API_definitions/quality-on-demand.yaml#L502)

Means that we only would like to provide information about the device, if this was explicitly provided by the user.
Or in other words: only when a 2-legged-token was used and the `device` is required to be in the request-body then we have to add the `device` also in the:

* `$.config.subscriptionDetail` of the response
* `$.data` in the cloud-event

If we are dealing with a 3-legged-token I aggree with @jlurien that we also have to remove the mandatory `device`-object from the CloudEvent `$.data`

dfischer-tech:
@jlurien @maxl2287 
Thank you for the notice – understood and accepted.

maxl2287 (approved):
LGTM

bigludo7 (approved):
LGTM

jlurien (approved):
@dfischer-tech , @maxl2287 The changes are OK, but we have to revert versions to wip and /vwip in the URL before merging (as in #303). At the same time, issue #302 could be fixed in this PR

maxl2287:
@jlurien @bigludo7 
Let's set back the versions via https://github.com/camaraproject/DeviceLocation/pull/305 first so we have a seperate PR for it.

jlurien:
> @jlurien @bigludo7 Let's set back the versions via #305 first so we have a seperate PR for it.

@dfischer-tech @maxl2287 already merged

PR #309, opened 2025-02-25
[geofencing-subscriptions]: Add "MISSING_IDENTIFIER" to the 422-error codes

### LINKED ISSUE #300: Missing error 422 with code MISSING_IDENTIFIER in Geofencing Subscriptions API
Problem description
In the specification "_geofencing-subscriptions.yaml_" the description part, code and example for "MISSING_IDENTIFIER" in representation for "CreateSubscriptionUnprocessableEntity422" is missing.
Missing error 422 with code MISSING_IDENTIFIER

Expected behavior
Add the missing error 422 with code MISSING_IDENTIFIER
Add in _info.description_ a part like:
```yaml
  # Identifying the device from the access token

    This API requires the API consumer to identify a device as the subject of the API as follows:
    - When the API is invoked using a two-legged access token, the subject will be identified from the optional `device` object, which therefore MUST be provided.
    - When a three-legged access token is used however, this optional identifier MUST NOT be provided, as the subject will be uniquely identified from the access token.

    This approach simplifies API usage for API consumers using a three-legged access token to invoke the API by relying on the information that is associated with the access token and was identified during the authentication process.

    ## Error handling:

    - If the subject cannot be identified from the access token and the optional `device` object is not included in the request, then the server will return an error with the `422 MISSING_IDENTIFIER` error code.

    - If the subject can be identified from the access token and the optional `device` object is also included in the request, then the server will return an error with the `422 UNNECESSARY_IDENTIFIER` error code. This will be the case even if the same device is identified by these two methods, as the server is unable to make this comparison.


```

Please also provide an example, like:
```yaml
            GENERIC_422_MISSING_IDENTIFIER:
              description: An identifier is not included in the request and the device or phone number identification cannot be derived from the 3-legged access token
              value:
                status: 422
                code: MISSING_IDENTIFIER
                message: The device cannot be identified.
```

maxl2287:
#### What type of PR is this?

* correction

Adds the `MISSING_IDENTIFIER` error in the 422 response.

Fixes #300 

#### Changelog input

```
 release-note
* add "MISSING_IDENTIFIER" to the 422 response of geofencing-subscriptions
```

### DIFF
--- a/code/API_definitions/geofencing-subscriptions.yaml
+++ b/code/API_definitions/geofencing-subscriptions.yaml
@@ -1345,6 +1345,7 @@ components:
                   code:
                     enum:
                       - IDENTIFIER_MISMATCH
+                      - MISSING_IDENTIFIER
                       - MULTIEVENT_SUBSCRIPTION_NOT_SUPPORTED
                       - SERVICE_NOT_APPLICABLE
                       - UNNECESSARY_IDENTIFIER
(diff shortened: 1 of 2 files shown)


bigludo7 (approved):
LGTM

jlurien (approved):
LGTM

jlurien:
@maxl2287 this could be merged. Thanks

PR #310, opened 2025-02-25
Location Retrieval - Align Error Test

### LINKED ISSUE #307: location-Retrieval: Align test plan with spec errors
**Problem description**
Commonalities 0.5 asks to have a test for all errors explicitly documented in the spec as relevant.

The following errors defined in location-retrieval.yaml are not explicitly tested by scenarios in location-retrieval.feature:

- 403 PERMISSION_DENIED
- 422 LOCATION_RETRIEVAL_MISSING_IDENTIFIER
- 429 QUOTA_EXCEEDED
- 429 TOO_MANY_REQUESTS

**Expected action**
Add test scenario if error is considered relevant or remove the error from spec.

- Add test for 403 as this error is mandatory
- Add tests for specific 422 Missing identifier test
- For location-retrieval, 429 is not specifically relevant. Implementations may return 429 for the usual cases even if it is not explicitly documented, so I'm in favor of removing it
- 404 NOT FOUND should be included along with 404 IDENTIFIER_NOT_FOUND in order to be allowed, although for this operation is not expected to be returned so we may remove it

**Alternative solution**

**Additional context**

bigludo7:
#### What type of PR is this?

* correction

- Remove 429 from the yaml
- add a summary for all examples in the yaml
- add a 403 Test Case in the feature file
- shift to 422 the Test Case for Missing Identifier 

Fixes #307 

404 NOT_FOUND is exceptionally kept in the spec even if there is no test for it

#### Changelog input

```
 release-note

```

#### Additional documentation 

This section can be blank.

```
docs

```

### DIFF
--- a/code/API_definitions/location-retrieval.yaml
+++ b/code/API_definitions/location-retrieval.yaml
@@ -190,8 +190,6 @@ paths:
           $ref: '#/components/responses/RetrieveLocationNotFound404'
         '422':
           $ref: '#/components/responses/RetrieveLocationUnprocessableEntity422'
-        '429':
-          $ref: '#/components/responses/Generic429'
       security:
         - openId:
             - location-retrieval:read
@@ -456,18 +454,21 @@ components:
                       - LOCATION_RETRIEVAL.MAXAGE_INVALID_ARGUMENT
           examples:
             GENERIC_400_INVALID_ARGUMENT:
+              summary: Invalid argument
               description: Invalid Argument. Generic Syntax Exception
               value:
                 status: 400
                 code: INVALID_ARGUMENT
                 message: Client specified an invalid argument, request body or query param.
             GENERIC_400_OUT_OF_RANGE:
+              summary: Out of range
               description: Out of Range. Specific Syntax Exception used when a given field has a pre-defined range or a invalid filter criteria combination is requested
               value:
                 status: 400
                 code: OUT_OF_RANGE
                 message: Client specified an invalid range.
             GENERIC_400_MAX_AGE_NOT_SATISFIABLE:
+              summary: Max age not satisfiable
               description: maxAge required in the request cannot be satisfied. Specific error for this API
               value:
                 status: 400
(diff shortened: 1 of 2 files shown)


maxl2287 (dismissed):

jlurien (commented):
I see these misalignments:

400 OUT_OF_RANGE -> defined but not tested. I don't think we need it looking into the request body schema
400 LOCATION_RETRIEVAL.MAXAGE_INVALID_ARGUMENT -> Description seems similar to 422 UNABLE_TO_FULFILL_MAX_AGE. is it needed? We should remove it or clarify when to use it instead of the other
422 LOCATION_RETRIEVAL.UNABLE_TO_FULFILL_MAX_SURFACE -> We may need a new scenario, something like

```  @location_retrieval_XX_location_retrieval_unable_to_locate_device_with_required_precision
  Scenario: Unable to provide device location with required maxSurface
    Given the testing device, identified by the token or provided in the request, is located within a surface of certain area
    And the request body property "$.maxSurface" is set to a value smaller than that area
    When the HTTP "POST" request is sent
    Then the response status code is 422
    And the response header "Content-Type" is "application/json"
    And the response header "x-correlator" has same value as the request header "x-correlator"
    And the response property "$.status" is 422
    And the response property "$.code" is "LOCATION_RETRIEVAL.UNABLE_TO_FULFILL_MAX_SURFACE"
    And the response property "$.message" contains a user friendly text

bigludo7:
Thanks @jlurien 

For the first 2 I need to update the yaml correct?
 400 OUT_OF_RANGE -> defined but not tested. I don't think we need it looking into the request body schema --> Agree I'm in favor to remove
400 LOCATION_RETRIEVAL.MAXAGE_INVALID_ARGUMENT -> Description seems similar to 422 UNABLE_TO_FULFILL_MAX_AGE. is it needed? We should remove it or clarify when to use it instead of the other --> Agree I'm in favor to remove 400 LOCATION_RETRIEVAL.MAXAGE_INVALID_ARGUMENT

Giving the purpose of the PR I propose to do this in another issue/PR - WDYT?

jlurien (approved):
LGTM

PR #314, opened 2025-02-27
Enhance description UNKNOWN vs UNABLE_TO_FULFILL_MAX_AGE

### LINKED ISSUE #308: Inconsistent behavior between Verify Location and Retrieve Location when a device is roaming
Inconsistent behavior between Verify Location and Retrieve Location when a device is roaming. 

In situations when the device is roaming it might not be possible to get its location. When this happens we have different behaviors.

**Verify Location** - It looks like two possible responses (and it is not obvious from the documentation which one will be responded)

- 200 with verification result UNKNOWN
- 422 LOCATION_VERIFICATION.UNABLE_TO_FULFILL_MAX_AGE "Unable to provide expected freshness for location"

**Retrieve Location** - It looks like two possible responses (and it is not obvious from the documentation which one will be responded)

- 404 LOCATION_RETRIEVAL.DEVICE_NOT_FOUND "The location server is not able to locate the mobile"
- 422 LOCATION_VERIFICATION.UNABLE_TO_FULFILL_MAX_AGE "Unable to provide expected freshness for location"

First thing to do is to document the expected behavior per API, that is when will you receive a 200/422 and 404/422.

Optionally we might want to revisit if we want to keep 200/422 for Verify Location or if we want to align with Retrieve Location and use 404/422 instead.

jlurien:
#### What type of PR is this?

* documentation

Enhance description to clarify usage after discussion in #308 

Partially resolves #308 for this meta
Issue to keep open

### DIFF
--- a/code/API_definitions/location-verification.yaml
+++ b/code/API_definitions/location-verification.yaml
@@ -17,10 +17,9 @@ info:
     * If the network's estimation of the device's location does not overlap with the requested area at all, the verification result is `FALSE`.
     * If the network's estimation of the device's location partially overlaps with the requested area, or it fully contains the requested area (because it is larger), the result is `PARTIAL`. In this case, a `match_rate` is included in the response, indicating an estimation of the likelihood of the match in percent.
     * Lastly, the network may not be able to locate the device. In this case, the verification result is `UNKNOWN`.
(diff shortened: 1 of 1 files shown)


maxl2287 (approved):
LGTM

bigludo7 (approved):
LGTM

PR #333, opened 2025-05-20
Polygon area type for geofences

### LINKED ISSUE #194: [Geofencing] Support for polygon area type

ilya-smirnov-berlin:
#### What type of PR is this?

* enhancement/feature

Geofences should support polygon type areas. Circle as only type is rarely used for geofencig. Usually a geofence describes a shape of an (often human built) geographical object like factories, buildings, city and country areas. 

Fixes #194 

Polygon types for location verification will be done in separate PR

#### Changelog input

```
 Optional support for polygon areas in geofencing subscriptions 
```

### DIFF
--- a/code/API_definitions/geofencing-subscriptions.yaml
+++ b/code/API_definitions/geofencing-subscriptions.yaml
@@ -12,7 +12,7 @@ info:
     * Legal restrictions regarding privacy, or other regulatory or implementation issues, may force the operator to set restrictions to the provided area, such as setting a minimum value to the accepted radius. In these cases, an error `422 GEOFENCING_SUBSCRIPTIONS.INVALID_AREA` will be returned and the error message will refer to the reason of the limitation.
     * If the provided polygonal type of area is not supported by operator, an error `422 GEOFENCING_SUBSCRIPTIONS.NOT_MANAGED_AREA_TYPE` will be returned.
 
-    The area provided in the request is described by a circle or a polygon. The circle area is determined by coordinates (latitude and longitude) and an accuracy defined by the radius. The polygon area is determined by an array of corner points in form of geo coordinates (latitude and longitude).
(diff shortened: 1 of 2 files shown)


bigludo7 (commented):
Thanks
- As discussed the polygon model should follow commonalities model as defined in lines 177-204 [here](https://github.com/camaraproject/Commonalities/blob/main/artifacts/CAMARA_common.yaml)
- Should we explicitly state that circle area-type management is mandatory for any API implementation? I guess yes.
- As Location verification is stable I tend to think that it should be handled in a separate PR. I'm in favor to be consistence between the API and to implement it for location-verification but probably easiest to have one PR per API.

ilya-smirnov-berlin:
Ready for review. Changes discussed in last meeting were applied.

ilya-smirnov-berlin:
Closed due to changed branch name. WIll be reopened with differnt PR

PR #357, opened 2025-06-25
[geofencing-subscriptions]: Define `id` to be mandatory in SubscriptionAsync

### LINKED ISSUE #355: The `id` in the asnychronous subscription-response needs to be mandatory
**Problem description**
The `id` is currently not mandatory.

**Expected behavior**
The `id` needs to be mandatory.

maxl2287:
#### What type of PR is this?

* enhancement/feature

Define `id` to be mandatory in SubscriptionAsync

Fixes #355 

#### Changelog input

```
 release-note
 * Define `id` to be mandatory in SubscriptionAsync
```

### DIFF
--- a/code/API_definitions/geofencing-subscriptions.yaml
+++ b/code/API_definitions/geofencing-subscriptions.yaml
@@ -659,6 +659,8 @@ components:
     SubscriptionAsync:
       description: Response for an event-type subscription request managed asynchronously (Creation or Deletion).
       type: object
+      required:
+        - id
       properties:
         id:
           $ref: "#/components/schemas/SubscriptionId"


code/API_definitions/geofencing-subscriptions.yaml
```diff
@@ -659,6 +659,8 @@ components:
     SubscriptionAsync:
       description: Response for an event-type subscription request managed asynchronously (Creation or Deletion).
       type: object
+      required:
```
jlurien, line 662: It seems to be still required: **https://github.com/camaraproject/Commonalities/blob/main/artifacts/camara-cloudevents/event-subscription-template.yaml#L551** am I looking in the wrong place?

maxl2287: @jlurien what do you mean?
I have added here the lines so that `id` is at the end here required.

It was not before

PR #368, opened 2025-08-20
Align radius with CAMARA_common.yaml

### LINKED ISSUE #364: Differences in Circle schema between APIs
**Problem description**

As reported [here](https://github.com/camaraproject/DeviceLocation/pull/360#issuecomment-3078149264):

Based on review script here is the comparison for `Circle ` schema

| File                          | Radius Type & Range              | Description Differences              |
|-------------------------------|----------------------------------|--------------------------------------|
| `location-verification.yaml`  | integer, min: 1, max: 200,000    | “Expected accuracy for verification…”|
| `location-retrieval.yaml`     | number, min: 1                   | “Distance from the center in meters” |
| `geofencing-subscriptions.yaml`| integer, min: 1, max: 200,000   | “Expected accuracy for subscription…”|

---

## Key Differences

  - `location-verification.yaml` and `geofencing-subscriptions.yaml` use `integer` type with a maximum value of 200,000.
  - `location-retrieval.yaml` uses `number` type (can be float) and  and does not specify a maximum (only minimum: 1).
- The description for `radius` differs slightly in context (verification, subscription, or generic).

---
Is there a reason for different definition of `radius` in `location-retrieval.yaml` ?
It can be discussed by the team in separate issue after M3.

**Possible evolution**

Discuss:
- should we change in location-retrieval from `number` to `integer`?
- is OK to keep `max` only in requests? 

`min` was reduced to 1 in requests after past discussions, with the clarification in description: "Note: The area surface could be restricted locally depending on regulations. Implementations may enforce a larger minimum radius (e.g. 1000 meters)."

jlurien:
#### What type of PR is this?

* correction

location-verification and geofencing-subscriptions used a schema for `radius` (property of `Area`, `Circle`) that was not aligned with model in CAMARA_common.yaml (already applied to location-retrieval)

Fixes #364 

Descriptions particularized for the APIs have been kept

#### Changelog input

```
radius for Circle defined as `number` without explicit maximum for `Location Verification` and Geofencing Subscriptions

```

### DIFF
--- a/code/API_definitions/geofencing-subscriptions.yaml
+++ b/code/API_definitions/geofencing-subscriptions.yaml
@@ -842,12 +842,11 @@ components:
             center:
               $ref: "#/components/schemas/Point"
             radius:
-              type: integer
+              type: number
               description: |
                 Expected accuracy for the subscription event of device location, in meters from `center`.
                 Note: The area surface could be restricted locally depending on regulations. Implementations may enforce a larger minimum radius (e.g. 1000 meters).
               minimum: 1
-              maximum: 200000
           required:
             - center
             - radius
--- a/code/API_definitions/location-verification.yaml
+++ b/code/API_definitions/location-verification.yaml
@@ -285,12 +285,11 @@ components:
             center:
               $ref: "#/components/schemas/Point"
             radius:
-              type: integer
+              type: number
               description: |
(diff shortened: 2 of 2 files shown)


bigludo7 (approved):
LGTM

maxl2287 (approved):
LGTM

PR #382, opened 2025-12-05
Enhance Area object with description

### LINKED ISSUE #379: What does the circle returned with the geofence event represent?
**Problem description**
In the geofence event notification data a mandatory area object is to be returned. From the documentation it is described as

- areaType - Type of this area. CIRCLE - The area is defined as a circle. 
- center - (Point)Coordinates (latitude, longitude) defining a location in a map.
- radius -  number >= 1
Expected accuracy for the subscription event of device location, in meters from center. Note: The area surface could be restricted locally depending on regulations. Implementations may enforce a larger minimum radius (e.g. 1000 meters).

This could be interpreted either as the geofence area used when creating the subscription or as the area in which the device is when the event is generated.

**Expected action**
- Discuss and decide if the area in the event is the geofence or the estimated location of the device. 
- Adjust the description in such a way that it is clear what this area represents 

**Additional context**
We also have an incorrect description of the purpose of the radius in the area object when creating a subscription. The radius is not really to set the "Expected accuracy for the subscription event of device location, in meters from center. " as written today but rather to set the border of the geofence. 

As touched upon in the discussions on #133 Define Guidelines for geofencing implementation the API invoker today lacks a way to set and/or understand the accuracy of a triggered event. Including the device location as a point with an accuracy could a potential mitigation.

bigludo7:
#### What type of PR is this?

* documentation

Add description for the geofencing area in the schema.

Fixes #379 

#### Changelog input

```
 release-note
- add description for area 

```

#### Additional documentation 

This section can be blank.

```
docs

```

### DIFF
--- a/code/API_definitions/geofencing-subscriptions.yaml
+++ b/code/API_definitions/geofencing-subscriptions.yaml
@@ -816,6 +816,7 @@ components:
 
     Area:
       type: object
+      description: The geofencing area where the monitor is active. This area is specified by API consumers in the subscription request. The same area definition is included in event notifications without any modifications.
       properties:
         areaType:
           $ref: "#/components/schemas/AreaType"


maxl2287 (approved):
LGTM, thanks @bigludo7 !

jlurien (approved):
LGTM

PR #386, opened 2026-02-09
Explicit minimum of 0 to maxAge

### LINKED ISSUE #385: Add minimum in spec to maxAge as it is specified to be >=0
**Problem description**
`maxAge` is specified just as integer in both location-verification and location-retrieval, but negative values must not be allowed, so it makes sense to explicitly define 0 as a minimum

**Expected behavior**
Add `minimum: 0`. This way any attempt to send a negative value will be covered by the test to return 400 INVALID_ARGUMENT when the input request does not comply with the request body schema

jlurien:
#### What type of PR is this?

* correction

Sets an explicit minimum of 0 to maxAge

Fixes #385 

#### Changelog input

```
Added `minimum: 0` for `maxAge` in location-verification and location-retrieval

```

### DIFF
--- a/code/API_definitions/location-retrieval.yaml
+++ b/code/API_definitions/location-retrieval.yaml
@@ -241,6 +241,7 @@ components:
           $ref: '#/components/schemas/Device'
         maxAge:
           type: integer
+          minimum: 0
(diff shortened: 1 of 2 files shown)


maxl2287 (approved):

bigludo7 (approved):
LGTM

PR #405, opened 2026-06-01
Update geofencing-subscriptions.yaml

jlurien:
#### What type of PR is this?

* correction

Fix error reported in #403


jlurien:
Dismissed

